PR #45, opened 2025-07-06
enhance: add gemini model

JimFuller-RedHat:
Allow default model to use gemini.

**Note:** at some point we will have more sophisticated model selection mechanism.

### DIFF
--- a/src/aegis/__init__.py
+++ b/src/aegis/__init__.py
@@ -9,6 +9,8 @@ import logging
 import os
 
 from dotenv import load_dotenv
+from pydantic_ai.models.anthropic import AnthropicModel
+from pydantic_ai.models.gemini import GeminiModel
 
 from rich.logging import RichHandler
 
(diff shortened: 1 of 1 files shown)


src/aegis/__init__.py
```diff
@@ -28,10 +30,15 @@
 
 tavily_api_key = os.getenv("TAVILY_API_KEY", "   ")
 
+# Simple logic for defining default model (we may have to make this more sophisticated).
 default_llm_model = OpenAIModel(
```
kdudka, line 34: I would move the `OpenAIModel` construction to the `else` branch so that it is skipped when it is not needed:
```diff
--- a/src/aegis/__init__.py
+++ b/src/aegis/__init__.py
@@ -31,14 +31,15 @@ llm_model = os.getenv("AEGIS_LLM_MODEL", "llama3.2:latest")
 tavily_api_key = os.getenv("TAVILY_API_KEY", "   ")

 # Simple logic for defining default model (we may have to make this more sophisticated).
-default_llm_model = OpenAIModel(
-    model_name=llm_model,
-    provider=OpenAIProvider(base_url=f"{llm_host}/v1/"),
-)
 if "api.anthropic.com" in llm_host:
     default_llm_model = AnthropicModel(model_name=llm_model)
-if "generativelanguage.googleapis.com" in llm_host:
+elif "generativelanguage.googleapis.com" in llm_host:
     default_llm_model = GeminiModel(model_name=llm_model)
+else:
+    default_llm_model = OpenAIModel(
+        model_name=llm_model,
+        provider=OpenAIProvider(base_url=f"{llm_host}/v1/"),
+    )


 @dataclass
```

PR #72, opened 2025-07-24
tests: enhance test suite to cache llm and tools

JimFuller-RedHat:


tests/README.md
```diff
@@ -0,0 +1,32 @@
...
+```
+
+To run a specific test:
+```commandline
+uv run pytest -k "test_suggest_impact_with_test_model"
+```
+
+To run a specific test with live llm/tooling:
+```commandline
+uv run pytest -k "test_suggest_impact_with_test_model"
```
kdudka, line 19: This is the same command as above.  Looks like a copy/paste mistake.

tests/README.md
```diff
@@ -0,0 +1,32 @@
...
+Note running 'live' will save results to LLM and osidb_caches.
+
+## Developing new tests
+
+When developing new tests or working on old tests one may have to recapture test data. 
+
+To recapture set env var:
+
+```commandline
+export TEST_RECAPTURE="true"
```
kdudka, line 31: The environment variable name sounds misleading to me.  I thought that it would fetch fresh data for all tests.  However, it only captures the data that has not yet been cached.  Something like `TEST_ALLOW_CAPTURE` would be more intuitive in my view.

tests/utils/osidb_cache.py
```diff
@@ -0,0 +1,44 @@
+import asyncio
```
kdudka, line 1: I would prefer to maintain the `osidb_cache` code at a single place.  We might consider also sharing the data, which comes with both advantages and disadvantages.  I can see that the content of `CVE-2025-0725.json` is different in each directory, I guess, because each of us captured the data using a different OSIDB instance.  In my case it was with `AEGIS_OSIDB_SERVER_URL="https://osidb-stage.prodsec.redhat.com/"`.

PR #73, opened 2025-07-25
chore: constrain input

JimFuller-RedHat:
We should ensure all LLM query input is constrained.

- [x] define common fields and use with feature input
- [x] add bad cve negative test
- [x] refactor osidb tools CVEID


JimFuller-RedHat:
eval test suite fails ... need to learn if we need to recapture, etc

PR #91, opened 2025-07-27
fix: make cwe more accurate

### LINKED ISSUE #71: `suggest-cwe-for-CVE-2025-23395` occassionally provides a wrong answer
The `suggest-cwe-for-CVE-2025-23395` evaluation case occasionally provides a wrong answer with high confidence rate, which results in the evaluation score being negative:
```
E       AssertionError: Unsatisfied assertion(s):
E       suggest-cwe-for-CVE-2025-23395: SuggestCweEvaluator(): score below threshold: -0.17515902199999966 < 0.1
```

The corresponding output looks like this:
```py
confidence=0.95
completeness=1.0
consistency=1.0
tools_used=['osidb_tool']
cve_id='CVE-2025-23395'
title='Local root exploit via `logfile_reopen()` in screen 5.0.0 with setuid-root bit set'
components=['screen']
explanation="The CVE describes a scenario where the 'screen' utility, running with setuid-root privileges, fails to drop these elevated privileges before handling a user-provided logfile path. This directly corresponds to CWE-273 (Improper Check for Dropping Privileges). Furthermore, because the program uses this user-supplied path to create a log file, it allows an attacker to control the file name and location, which aligns with CWE-73 (External Control of File Name or Path). The combination of these weaknesses allows an unprivileged user to create a root-owned file in an arbitrary location, leading to privilege escalation."
cwe=['CWE-273', 'CWE-73']
```

- [CWE-273](https://cwe.mitre.org/data/definitions/273.html) means `The product attempts to drop privileges but does not check or incorrectly checks to see if the drop succeeded.`
- [CWE-73](https://cwe.mitre.org/data/definitions/73.html) means `The product allows user input to control or influence paths or file names that are used in filesystem operations.`

None of the above CWEs really describes the situation where the program simply does not attempt to drop the privileges where it should.  The most accurate suggestions (in the order of preference) would be:
- [CWE-271](https://cwe.mitre.org/data/definitions/271.html), which means `The product does not drop privileges before passing control of a resource to an actor that does not have those privileges.`
- [CWE-250](https://cwe.mitre.org/data/definitions/250.html), which means `The product performs an operation at a privilege level that is higher than the minimum level required, which creates new weaknesses or amplifies the consequences of other weaknesses.`
- [CWE-272](https://cwe.mitre.org/data/definitions/272.html), which means `The elevated privilege level required to perform operations such as chroot() should be dropped immediately after the operation is performed.`

JimFuller-RedHat:
fixes #71 

Note: There is a limit to how accurate we can make accurate analysis with model choice, configuration or RAG ... in this instance it seems to have worked but better accuracy will eventually require model fine tuning (or at worst training).

### DIFF
--- a/src/aegis_ai/agents.py
+++ b/src/aegis_ai/agents.py
@@ -31,8 +31,9 @@ class AegisAgent(Agent):
             model=default_llm_model,
             model_settings=default_llm_settings
             | {
-                "temperature": 0.05,
+                "temperature": 0.06,
                 "top_p": 0.8,
+                "seed": 42,
                 "response_format": {"type": "json_object"},
             },
             **kwargs,
--- a/src/aegis_ai/features/cve/__init__.py
+++ b/src/aegis_ai/features/cve/__init__.py
@@ -63,10 +63,10 @@ class SuggestCWE(Feature):
 
     async def exec(self, cve_id: CVEID):
         prompt = AegisPrompt(
-            user_instruction="Your task is to meticulously examine the provided CVE JSON object and suggest CWE-ID(s) for the CVE",
-            goals="""
-                Given a description of a Common Vulnerabilities and Exposures (CVE), accurately predict its Common Weakness Enumeration (CWE). List all CWE that are applicable.
-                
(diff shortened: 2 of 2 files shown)


JimFuller-RedHat:
eval pass

PR #108, opened 2025-07-31
Fix ValueError message

hemajv:
small fix for printing correct value error messages


JimFuller-RedHat (approved):
nice catch, we also have to add aegis specific error abstractions at some point - LGTM

PR #114, opened 2025-08-04
chore: add openapi.yml

JimFuller-RedHat:
- [x] add docs/openapi.yml
- [x] add dynamic /openapi.yml to Aegis REST-API
- [x] add web test


C-Valen (commented):
Should we also add a link on the `index.html` for the yml file, just as there is for json?
Something like:
`<a href='/openapi.yml'>openapi yaml</a><br/>`

JimFuller-RedHat:
@C-Valen good idea, added link - eventually the web pages will get some sort of style treatment

PR #117, opened 2025-08-04
feature: example of prompt safety

JimFuller-RedHat:
This is mostly another experiment helping define scope for agentic workflow ... in this instance we will always want to check prompt 

```
AEGIS_SAFETY_ENABLED="true"
AEGIS_SAFETY_LLM_HOST="https://granite3-guardian-2b.somellmhost.com:443"
AEGIS_SAFETY_LLM_MODEL="granite3-guardian-2b"
AEGIS_SAFETY_OPENAPI_KEY="someopenapikey"
```

we could do this as a step/node in agentic graph/ workflow, or as a method on prompt object.


src/aegis_ai/__init__.py
```diff
@@ -70,7 +65,25 @@ class AppSettings(BaseSettings):
...
     default_llm_settings: Dict = default_llm_settings
-    otel_enabled: bool = otel_enable
+
+    otel_enabled: bool = os.getenv("AEGIS_OTEL_ENABLED", "false").lower() in (
+        "true",
+        "1",
+        "t",
+        "y",
+        "yes",
+    )
```
kdudka, line 75: It would be good to define the tuple at a single place.

src/aegis_ai/prompt.py
```diff
@@ -82,6 +85,30 @@ class AegisPrompt(BaseModel):
...
+            <start_of_turn>
+            User Message: {self.to_string()}
+            <end_of_turn>
+    
+            Our safety risk definition is defined below:
+    
+            <start_of_risk_definition>
+            * {risk_definition}
+            <end_of_risk_definition>
+
```
kdudka, line 106: All the blank lines in this string are misaligned.

src/aegis_ai/__init__.py
```diff
@@ -70,7 +65,25 @@ class AppSettings(BaseSettings):
...
+
+    safety_enabled: bool = os.getenv("AEGIS_SAFETY_ENABLED", "false").lower() in (
```
kdudka, line 77: I would insert a comment to explain that this only takes an effect if `prompt.is_safe()` is explicitly called.

kdudka (approved):

PR #128, opened 2025-08-07
chore: update pydantic-ai to 0.6.0

kdudka:
... which is the latest upstream release.  There seem to be no incompatibilities at first glance.

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -237,7 +237,7 @@ wheels = [
 
 [[package]]
 name = "anthropic"
-version = "0.54.0"
+version = "0.61.0"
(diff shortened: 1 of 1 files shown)


JimFuller-RedHat (approved):
lgtm

JimFuller-RedHat:
I think we need to update pyproject.toml as well

kdudka:
> I think we need to update pyproject.toml as well -usually uv add does that

That is needed when `pyproject.toml` pins a dependency to a specific version.  But for `pydantic-ai` we have `>=0.4.11`, which would install the latest available version (unless there is a conflicting relation with other deps).

PR #137, opened 2025-08-20
make the `xgb-training` experiment easier to reproduce

kdudka:
- Do not process only 5 CVEs by default.

- Remove CVEs from the training set if they were in the testing set, too.  These sets should be disjoint.

- Import summarized data for the whole data set rather than for the first 5 CVEs.

- Update the split data sets and the balanced training set.  Some CVEs were missing in the data set.  Some CVEs ended up with different boolean values assigned to features (probably implied by the recent code changes).

- Re-run the training and evaluation with the updated data.  The accuracy is now slightly better.

- Improve handling of per CVE timeout `cve_feature_extraction.py`.  If the process was signalled, the processing needs to be interrupted because no additional `SIGALRM` signal will be received for the specific CVE.  This helps to get the results in a predictable amount of time.

- Parallelize the feature extraction based on the number of available CPUs.

Related: https://issues.redhat.com/browse/AEGIS-111


kdudka:
@suppathak Thanks for review!

The updated  evaluation results for the full set of CVEs is imported in commit 87c31255d0610b16ae6c57bf9e7d6a1a99cf0146.

If you [scroll down](https://github.com/RedHatProductSecurity/aegis-ai/pull/137/commits/87c31255d0610b16ae6c57bf9e7d6a1a99cf0146#diff-c76aa2407ed30d223f0b7208b1a4c1e2a730739ebb3c53cf62a0be12e8bdb98e) in the commit view, GitHub will show you side-by-side the old and new version of `test_results.png`.

PR #143, opened 2025-08-25
fix: simplify prompts for smaller LLMs [WIP]

kdudka:
Related: https://issues.redhat.com/browse/AEGIS-110

### DIFF
--- a/src/aegis_ai/features/cve/__init__.py
+++ b/src/aegis_ai/features/cve/__init__.py
@@ -22,38 +22,19 @@ class SuggestImpact(Feature):
 
     async def exec(self, cve_id: CVEID, static_context: Any = None):
         prompt = AegisPrompt(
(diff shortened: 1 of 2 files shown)


kdudka:
> happy to see evals passing with mistral-small-24b-w8a8

They are passing only partially for now.  Only 2-3 tests from 6 are usually green (you can see it in the CI log) and we take it as success.  This behavior can be controlled by the `AEGIS_EVALS_MIN_PASSED` GitHub variable.  The main obstacle for more tests passing is currently the LLM that we use for `LLMJudge`, which can be controlled separately.  The LLM that we use now fails to verify simple assertions, as for example:
```
suggest-impact-for-CVE-2024-53232: name='LLMJudge' arguments={'rubric': "The 'explanation' output field does not list affected Red Hat products.  Ignore other fields in the output.", 'model': 'openai:mistral-small-24b-w8a8', 'model_settings': {}}: The output lists affected Red Hat products: 'affected_products':[],
suggest-impact-for-CVE-2025-5399: name='LLMJudge' arguments={'rubric': "The 'explanation' output field does not list affected Red Hat products.  Ignore other fields in the output.", 'model': 'openai:mistral-small-24b-w8a8', 'model_settings': {}}: The 'explanation' field does not mention affected Red Hat products
```

PR #174, opened 2025-09-11
Add minimal dockerfile

JimFuller-RedHat:
- [x] defined Containerfile
- [x] tweaked uvicorn settings
- [x] some minor tweaks

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,42 @@
+.attic/
+docs/
+tests/
+evals/
+etc
+src/aegis_ai_ml
+src/aegis_ai_chat
+
+
+# Environment
+.venv/
+env/
+venv/
+
+# IDE and OS
+.vscode/
+.idea/
+*.sublime-*
+Thumbs.db
+.DS_Store
+
+# Test and coverage
+.pytest_cache/
+.coverage
+htmlcov/
+.ruff_cache/
+*.pytest_cache*
+
+# Local data
+*.log
+
+# Packaging
+dist/
+build/
+*.egg-info/
+
+# Git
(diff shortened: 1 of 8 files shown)


pyproject.toml
```diff
@@ -25,15 +25,18 @@ dependencies = [
     "packageurl-python>=0.16.0",
     "platformdirs>=4.3.8",
     "pydantic>=2.11.7",
-    "pydantic-ai[logfire]>=0.0.1",
+    "pydantic-ai>=0.0.1",
     "pydantic-ai-slim[duckduckgo,mcp,tavily]>=1.0.1",
     "pygithub>=2.7.0",
+    "requests-gssapi>=1.3.0",
```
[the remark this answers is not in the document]
JimFuller-RedHat: I believe one of the tools needed it - I will separate out

.dockerignore
```diff
@@ -0,0 +1,43 @@
...
+# Local data
+*.log
+
+# Packaging
+dist/
+build/
+*.egg-info/
+
+# Git
+.git
```
kdudka, line 38: @JimFuller-RedHat This breaks the CI because `Containerfile.eval` explicitly copies `.git` into the container image.  This is used to create optimized evaluation images that do not need to clone the repository at run time.  They only fetch the changes done since the image creation if needed.

kdudka: I have pushed commit 1ad4998abfdd26faa61628e48a9b3676080a9d23 to fix this.

kdudka (approved):
I have pushed two fixup commits.  Please feel free to squash them.

PR #190, opened 2025-09-17
chore: define a pull request template for Aegis

kdudka:
If the template is used by the contributors, we will record information that will be useful later on.

This template will be used for pull requests opened from now on.

Related: https://issues.redhat.com/browse/AEGIS-141

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,7 @@
+## What
+
+## Why
+
+## How to Test
+
+## Related Tickets


.github/pull_request_template.md
```diff
@@ -0,0 +1,7 @@
+## What
+
+## Why
+
+## How to Test
+
+## Related Tickets
```
iswiac, line 7: If it is a Jira ticket, than may be worth to add "JIRA". 

Also not related to this remplate, but would be great if we have an automatic JIRA comment added indicating what pull requests are associated with that specific JIRA.

PR #212, opened 2025-09-25
tool: add intentionally simple http get tool

JimFuller-RedHat:
## What

Adds a very simple http get tool which aegis agent can use to dereference urls it finds.

## Why

Specifically we want aegis to be able to retrieve additional context from reference urls.

## How to Test

```
uv run aegis search "use http_get tool to retrieve this web page https://en.wikipedia.org/wiki/Open-world_assumption return first paragraph"
```


kdudka (changes requested):
I am concerned about the security implications of this tool.  While it might be useful for some initial experiments, I do not think this should be enabled in the production or stage environment, or even on the Red Hat internal network as such.

Also the `SEC-NET-REQ-5` ESS control requires the outgoing Internet traffic to be more restricted and persistently logged:
> All outgoing network connections (to internet) originated from applications/databases/systems/platforms that are deployed in Red Hat data centers or at cloud providers where we have administrative control must:
>
> 1). log the connection details and those logs must be sent to enterprise log collection & analyzing platform (eg Splunk)
> 2). be prevented from making connections to potential rogue sites or hosts by consuming Infosec approved "deny" lists (domains/sites/IPs) via DNS RBLs or other related Infosec approved services/tools.
> 3). must enforce "allow" list egress network connections to the public internet. Application owners are responsible for pre-authorizing or providing the list of "allow" list of all internet-based resources that their applications/systems need to connect to (such as GitHub, Salesforce APIs, etc)

JimFuller-RedHat:
> I am concerned about the security implications of this tool. While it might be useful for some initial experiments, I do not think this should be enabled in the production or stage environment, or even on the Red Hat internal network as such.
>

Stepping back, our usage of public models like gemini have exactly the same implications eg.  'get gemini to goto evil server' and how to mitigate possible evil context injected into analysis. This tool (apart from unknown bugs) does not increase existing security surface area.

I think with **allow_list** this can be used in production against well known and we use on a small handful of well known security information public sites - this is especially useful for getting up to date information that has not yet been crawled by search engines (ex. seen by tavily tool) yet referenced by vulnerability.

src/aegis_ai/tools/http_get/__init__.py
```diff
@@ -0,0 +1,116 @@
...
+            return URLContent(
+                requested_url=url,
+                status="unsupported_method",
+                error_message="URL must start with 'http://' or 'https://'.",
+            )
+
+        parsed_url = urlparse(url)
+        hostname = parsed_url.netloc.lower()
+        for allowed_host in allow_list:
+            if allowed_host not in hostname:
```
kdudka, line 86: If `github.com` is in `allow_list`, would this check allow a connection to `github.com.evil-domain.ru`?

src/aegis_ai/tools/http_get/__init__.py
```diff
@@ -0,0 +1,116 @@
...
+                logger.warning(
+                    f"http_get tool is not permitted to access domain: '{hostname}'."
+                )
+                return URLContent(
+                    requested_url=url,
+                    status="excluded_domain",
+                    error_message=f"Domain '{hostname}' is not permitted.",
+                )
+
+        response = requests.get(
```
kdudka, line 96: Should we limit the size of received/processed data somewhere?

JimFuller-RedHat: not just yet, I have a general task (across all tools) related to context token size (which relates to this) and usage and retries ... which should be done as part of that PR.

src/aegis_ai/__init__.py
```diff
@@ -78,6 +78,7 @@
 use_nvd_dev_tool = os.getenv("AEGIS_USE_MITRE_NVD_MCP_TOOL_CONTEXT", "false")
 use_dbpedia_tool = os.getenv("AEGIS_USE_DBPEDIA_TOOL_CONTEXT", "false")
 use_cisa_kev_tool = os.getenv("AEGIS_USE_CISA_KEV_TOOL_CONTEXT", "false")
+use_http_get_tool = os.getenv("AEGIS_USE_HTTP_GET_TOOL_CONTEXT", "false")
```
kdudka, line 81: This value does not seem to be used anywhere.

src/aegis_ai/tools/http_get/__init__.py
```diff
@@ -0,0 +1,158 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+def compile_allowed_hosts_regex() -> Optional[re.Pattern]:
+    """
+    Loads AEGIS_HTTP_GET_TOOL_ALLOW_LIST from env, which is a comma-delimited list in a string.
+    Escapes non-regex characters where necessary, and compiles patterns into a single, case-insensitive
+    regex using OR operator (|).
+    """
```
kdudka, line 24: Honestly, this looks wild.  I think we should clearly mark this as an experimental feature.  If we ever want to support this as a security feature, we should go for a syntax that is already being used elsewhere, such as the `network.negotiate-auth.trusted-uris` configuration field in Firefox, or `AuthServerAllowlist` from Google Chrome (and ideally reuse some already trusted solution for this).  Otherwise this is going to be a frequent source of security issues.

PR #218, opened 2025-09-27
suggest-cwe: ensure feature has access to all cwe definitions for ana…

JimFuller-RedHat:
## What

Provide **suggest-cwe** the entire corpus of allowed cwes before we start analysing.

## Why

This helps inform analysis and has positive impact on suggest-cwe eval test.

## How to Test

LOGFIRE_IGNORE_NO_CONFIG=1 uv run pytest -vv -s evals/features/cve/test_suggest_cwe.py


src/aegis_ai/tools/cwe/__init__.py
```diff
@@ -4,34 +4,43 @@
...
     "https://cwe.mitre.org/data/csv/699.csv.zip",  # development - the only view supported by OSIM
     # "https://cwe.mitre.org/data/csv/1000.csv.zip",  # research
     # "https://cwe.mitre.org/data/csv/1008.csv.zip",  # architectural
     # "https://cwe.mitre.org/data/csv/1081.csv.zip",  # entries with maintenance notes
 ]
 
+# lazy load CWE
+CACHE_DIR = f"{config_dir}/mitre_cwe"
+CACHE_FILE = "cwe_full_defs.json"
+os.makedirs(CACHE_DIR, exist_ok=True)
```
kdudka, line 35: @JimFuller-RedHat Do we really want to create directories on module import?  That feels unexpected to me.

JimFuller-RedHat: happy to  move the call to whenever we persist the cache - makes sense.

PR #234, opened 2025-10-03
fix: do not hard-code project version at multiple places

kdudka:
## What
The version now reflects the output of `git describe`.  When a stable release is produced, a fixed version string is embedded in the dist package.

TODO

 [ ] update the version string even when `pyproject.toml` is not touched
 [ ] update Konflux CI to create/update container image tag when it is
     built from a tagged commit

## Why
TBD

## How to Test
TBD

## Related Tickets
Related: https://issues.redhat.com/browse/AEGIS-31


JimFuller-RedHat:
we can use tag we use for github release eg. propose when this is done that I create a new github/pypi release and we can use that tag for prod deployment.

PR #242, opened 2025-10-08
fix: make the container image builds work again

kdudka:
## What
fix: make the container image builds work again

## Why
As CPython 3.14.0 became available in the build environment, the builds of Aegis started to fail:
```
Downloading cpython-3.14.0-linux-x86_64-gnu (download) (33.1MiB)
 Downloading cpython-3.14.0-linux-x86_64-gnu (download)
Using CPython 3.14.0
Creating virtual environment at: .venv
Resolved 253 packages in 3.06s
error: Distribution `torch==2.8.0+cpu @ direct+https://download.pytorch.org/whl/cpu/torch-2.8.0%2Bcpu-cp313-cp313-manylinux_2_28_x86_64.whl` can't be installed because the binary distribution is incompatible with the current platform
```

This commit enforces use of CPython 3.13.x to make the build work again.

## How to Test
The CI is green again.

## Related Tickets
Related: https://issues.redhat.com/browse/AEGIS-198

## Summary by Sourcery

Restrict the supported Python version to 3.13.x to avoid incompatible PyTorch binaries on 3.14 and restore successful container image builds

Enhancements:
- Constrain `requires-python` in pyproject.toml to "<3.14"
- Regenerate uv.lock to reflect the updated dependency constraints


JimFuller-RedHat (approved):
LGTM

PR #308, opened 2025-11-03
enhance: read the `AEGIS_CORS_TARGET_REGEX` env var

kdudka:
## What
Read the `AEGIS_CORS_TARGET_REGEX` env var.

Also do not accept `https://localhost` and `https://localhost:5173` when `AEGIS_CORS_TARGET_REGEX` is set to a regex that does not match these URLs.

## Why
So that one can specify multiple URLs as CORS origins, such as:
```
export AEGIS_CORS_TARGET_REGEX='https://osim-(stage|uat)\.example\.com'
```

## How to Test
Specify multiple CORS target URLs as outlined above.

## Related Tickets
Related: https://issues.redhat.com/browse/AEGIS-230

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - simplified `AegisFeatureModel` to reduce LLM overhead
 - refactored agents and toolsets internals
 - renamed `rewrite-{description,statement}` to `suggest-{description,statement}`, respectively
+- `AEGIS_CORS_TARGET_URL` was replaced by `AEGIS_CORS_TARGET_REGEX` to support multiple CORS origins
 
 ### Added
 - added `/healthz` endpoint of the web server without authentication and logging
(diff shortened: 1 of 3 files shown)


JimFuller-RedHat (approved):
maybe consider escaping the regex to make it safer, otherwise LGTM

JimFuller-RedHat:
> > maybe consider escaping the regex to make it safer, otherwise LGTM
> 
> You mean to escape the special characters in the regex? Then we would not take a regex any more...
> 
> Perhaps we could rename the variable to `AEGIS_CORS_TARGET_REGEX`, as suggested by @sourcery-ai, to make it more obvious (at the cost of backward incompatibility and the need for a more coordinated upgrade).

yes even better ;)

kdudka:
@JimFuller-RedHat I have further simplified the configuration and made it more secure at the same time:
- If `AEGIS_CORS_TARGET_REGEX` is unset, the following URLs are accepted (same as before this change):
    - `http://localhost:5173`
    - `https://localhost`
    - `https://localhost:5173`

- If `AEGIS_CORS_TARGET_REGEX` is set, only URLs matching the specified regex are allowed.

PR #317, opened 2025-11-07
enhance: log the outcome of `suggest-impact` feature

kdudka:
## What
- evals: add a `suggest-cwe` case based on feedback
- update pinned dependencies in `uv.lock`
- decrease verbosity of `search_cwes` when not debugging
- introduce a logging interface for `AegisFeatureModel`
- log the outcome of `suggest-impact` feature

## Why
To get more useful logs both from evaluation and from staging/prod Aegis instances.

## How to Test
See the CI results.

## Related Tickets
Related: https://issues.redhat.com/browse/AEGIS-65

## Summary by Sourcery

Introduce a logging interface for feature models and use it to log the outcomes of suggest-impact and suggest-CWE features, adjust log levels for CWE search, add a new evaluation test case, and update pinned dependencies.

Enhancements:
- add printable_outcome hook to AegisFeatureModel and override in SuggestImpactModel and SuggestCWEModel to enable outcome logging
- modify run_if_safe to log feature outcomes using the new printable_outcome method
- reduce verbosity of CWE search logs by changing matches from info to debug level

Build:
- update pinned dependencies in uv.lock

Tests:
- add a SuggestCWE evaluation case for CVE-2022-50558

### DIFF
--- a/src/aegis_ai/features/cve/data_models.py
+++ b/src/aegis_ai/features/cve/data_models.py
@@ -71,6 +71,10 @@ class SuggestImpactModel(AegisFeatureModel):
         description="Suggested Red Hat CVSS3.1 vector",
     )
 
+    def printable_outcome(self) -> str:
+        """override the logging hook to print the resulting suggestion"""
+        return f"{self.impact} {self.cvss3_score} {self.cvss3_vector}"
+
 
 class SuggestCWEModel(AegisFeatureModel):
     """


JimFuller-RedHat (approved):
LGTM

kdudka:
> The base printable_outcome returns a generic "[...]" which could lead to unhelpful logs if a feature doesn’t override it—consider making it abstract or providing a default that serializes the model’s key attributes.

This is intentional to avoid leaking sensitive data into logs.  For full-text suggestions, we might need to hide the answers in the log for embargoed flaws.

PR #345, opened 2025-11-15
osidb-tool:  enable dynamic filtering based on data dep injection

JimFuller-RedHat:
## What

- [x] dynamically filter CVE fields returned by `osidb_tool` based on feature (using [data injection](https://ai.pydantic.dev/dependencies/))
- [x] we also need to dynamically filter CVE fields when static content is supplied (as we do direct from `OSIM`)
- [x] refactored app settings
- [x] added `AEGIS_LLM_TEMPERATURE`,`AEGIS_LLM_TOP_P` and `AEGIS_LLM_MAX_TOKENS` env var.
- [x] bumped pydantic-ai 1.20.0

## Why

Today we do not include top level CVE impact and cwe-id into osidb CVE information ... to avoid any bias. We should be able to manually filter such data instead of rely on 'prompt-only' constraints.

Cleaning up the app settings ensure we consistently set model (and their settings) correctly everywhere.

## How to Test

Run any eval test:
```
uv run pytest -vv -s evals/features/cve/test_suggest_impact.py
```

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 - bump to pydantic-evals,pydantic-ai 1.20.0
+- dynamic filtering of CVE data - using data deps injection with `osidb_tool`.
+- dynamic filtering of CVE data when supplied direct with static content.
 
 
 ## [0.4.2] - 2025-11-14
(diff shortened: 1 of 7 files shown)


kdudka (changes requested):
I am still unable to make this work.  The failures do not seem to be related to evals or `osidb_cache`.

When I run `uv run aegis cvss-diff CVE-2024-53232` on the `main` branch, it completes in 16s of wall-clock time.

When I run the same command on the `minor-settings-refactor` branch, it takes 76s to fail with `pydantic_ai.exceptions.UnexpectedModelBehavior: Exceeded maximum retries (5) for output validation`.

PR #627, opened 2026-05-11
chore: consolidate runtime dependencies

superbuggy:
## What

Move `pandas>=1.5.0` and `imbalanced-learn>=0.13.0` from `ml_deps` into `classifier_deps`, and remove the duplicate entries from `ml_deps` (they're now inherited transitively via `aegis-ai[classifier_deps]`).

## Why

The production container image installs `--extra classifier_deps` and needs both packages for XGBoost retraining (`make retrain-kernel`). Previously, these were only in `ml_deps` (which also pulls torch, transformers, and other SecBERT libraries). Adding them to `classifier_deps` raises the production image from ~1.93 GB to ~2.03 GB. Since the Lead Engineer concluded that a 100 MB delta doesn't justify maintaining a separate training image, consolidating to one image that handles both inference and retraining is the pragmatic choice. The classifier remains optional at runtime via `AEGIS_USE_KERNEL_CLASSIFIER`.

## How to Test

- [ ] `uv sync --extra classifier_deps` resolves without errors
- [ ] `make retrain-kernel` completes successfully (verifies training scripts find pandas and imbalanced-learn)
- [ ] `make build-container` builds and produces ~2.03 GB image
- [ ] Existing test suite passes (`make all`)

## Related Tickets
https://redhat.atlassian.net/browse/AEGIS-423

## Summary by Sourcery

Consolidate runtime dependency groups to ensure classifier-related packages are available via the classifier_deps extra while avoiding duplication.

Build:
- Move pandas and imbalanced-learn from ml_deps to classifier_deps so the classifier_deps extra fully covers classifier retraining needs without duplicating dependencies in ml_deps.

Chores:
- Regenerate the uv lockfile to reflect the updated optional dependency groups.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -132,6 +132,8 @@ Issues = "https://github.com/RedHatProductSecurity/aegis-ai/issues"
 
 [project.optional-dependencies]
 classifier_deps = [
+    "imbalanced-learn>=0.13.0",
+    "pandas>=1.5.0",
     "scikit-learn>=1.1.0",
     "xgboost>=1.7.0",
 ]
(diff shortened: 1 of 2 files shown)


kdudka (approved):
Looks good.  Thanks!

PR #630, opened 2026-05-12
fix: do not suggest `kernel` component for `macOS` CVEs

kdudka:
## What
Teach `SuggestAffectedComponents` to use the OS name (e.g. 'macOS') instead of 'kernel' for proprietary OS kernel vulnerabilities.

## Why
We use `kernel` for the Linux kernel only.

## How to Test
See the evals results.

## Related Tickets
Related: https://redhat.atlassian.net/browse/AEGIS-404

## Summary by Sourcery

Adjust CVE component suggestion rules for proprietary OS kernel vulnerabilities and extend evaluation coverage for a new macOS CVE.

Documentation:
- Clarify that proprietary OS kernel vulnerabilities (e.g. macOS, iOS, Windows) should use the OS name as the affected component rather than 'kernel', reserving 'kernel' for the Linux kernel.

Tests:
- Add an eval case for CVE-2026-28925 to validate affected component suggestions for a proprietary OS kernel vulnerability.


superbuggy (approved):

superbuggy:
I wonder if makes sense to do something similar for FreedBSD, as well

kdudka:
> I wonder if makes sense to do something similar for FreedBSD, as well

Thanks for heads up!  I checked OSIDB stage/prod and so far we do not have any case where `FreeBSD` was in `comment_zero` and `kernel` in `components` at the same time.

PR #633, opened 2026-05-12
fix: drop misleading cvss ignore clause from suggest-impact prompt

superbuggy:
## What

Removed the "Ignore any pre-labeled impact/CVSS and decide independently." clause from the `SuggestImpact` LLM prompt in `src/aegis_ai/features/cve/__init__.py`.

## Why

The clause was misleading: `feature_deps(exclude_osidb_fields=["impact", "rh_cvss_score"])` already strips Red Hat's internal impact label and CVSS score from the context before the prompt is built, so instructing the model to "ignore" fields that aren't present is inconsistent. Worse, the wording could inadvertently signal the LLM to also disregard external collector CVSS scores (e.g. NVD), which are intentionally kept in context. Fixes AEGIS-412.

## How to Test

Run the suggest-impact eval suite and confirm no regressions:

```bash
uv run pytest evals/features/cve/test_suggest_impact.py -x
uv run pytest evals/features/cve/test_suggest_impact_kernel_cves.py -x
```

## Related Tickets
https://redhat.atlassian.net/browse/AEGIS-412

## Summary by Sourcery

Align the suggest-impact LLM prompt with available context by removing an instruction to ignore pre-labeled impact/CVSS data that is no longer present.

### DIFF
--- a/src/aegis_ai/features/cve/__init__.py
+++ b/src/aegis_ai/features/cve/__init__.py
@@ -448,7 +448,7 @@ class SuggestImpact(Feature):
             )
 
         prompt = AegisPrompt(
-            user_instruction="Analyze the CVE JSON and derive a CVSS v3.1 base vector and score with metric-by-metric rationale from the perspective of Red Hat customers. Based on the score, select the impact (LOW/MODERATE/IMPORTANT/CRITICAL). Ignore any pre-labeled impact/CVSS and decide independently.",
+            user_instruction="Analyze the CVE JSON and derive a CVSS v3.1 base vector and score with metric-by-metric rationale from the perspective of Red Hat customers. Based on the score, select the impact (LOW/MODERATE/IMPORTANT/CRITICAL).",
             goals="""
                 - Return exactly one CVSS:3.1 base vector and score consistent with each other.
                 - Provide short reasoning for each base metric (AV, AC, PR, UI, S, C, I, A).


kdudka:
@superbuggy I can see some `suggest-impact` evaluation cases failing consistently.  This does not happen on the `main` branch.

superbuggy:
> @superbuggy I can see some `suggest-impact` evaluation cases failing consistently. This does not happen on the `main` branch.

It looks like no tools were invoked for CVE-2025-40320 (`tools_used=[]`) which is fixed by #637. Quite odd that the prompt change is causing this failure repeatedly, but perhaps is a side effect of the fixed seed.

superbuggy:
>It looks like no tools were invoked for https://github.com/advisories/GHSA-2vg8-pj5q-762f (tools_used=[]) which is fixed by https://github.com/RedHatProductSecurity/aegis-ai/pull/637. Quite odd that the prompt change is causing this failure repeatedly, but perhaps is a side effect of the fixed seed.

I ran into similar errors on evals #637, and dug in further and this does not seem to be the cause. Analysis is below and in https://github.com/RedHatProductSecurity/aegis-ai/pull/637#issuecomment-4447531881

<details>
<summary>Show analysis</summary>

## Failure 1 — CVE-2023-53764: `Exceeded maximum output retries (1)` (root cause only)

### Root Cause Chain

The error is pydantic-ai's `UnexpectedModelBehavior` raised when the Gemini 2.5 Flash model fails `SuggestImpactModel` schema validation twice (initial + 1 retry). Five layers contribute:

**Layer 1: Google API blocks schema-constrained generation when function tools are present.**
In pydantic-ai's Google adapter ([`_build_content_and_config`](https://github.com/pydantic/pydantic-ai)), when `output_mode == 'native'` and `function_tools` are present, it raises `UserError`. So pydantic-ai falls back to `'prompted'` mode, where the JSON schema is added to the system prompt and `response_mime_type='application/json'` is set -- but no `response_schema` constraint. The model produces free-form JSON that must pass post-hoc validation.

**Layer 2: `SuggestImpactModel` has strict validators that are easy to trip without constrained decoding.**
- `cvss3_vector: Optional[CVSS3Vector]` -- `BeforeValidator` calls `cvss.CVSS3(v)`, rejects any typo or missing metric
- `disclaimer: Literal["This response was generated by Aegis AI..."]` -- must reproduce a 174-character sentence exactly, character-for-character
- `confidence: float` with `ge=0.0, le=1.0` bounds
- `impact: Optional[Literal["LOW", "MODERATE", "IMPORTANT", "CRITICAL"]]` -- case-sensitive

Any of these failing causes pydantic-ai to send the validation error back to the model and retry.

**Layer 3: `_max_output_retries = 1` (confirmed).**
`create_aegis_agent` does not set `output_retries`, so pydantic-ai defaults to `retries=1`. The model gets exactly 2 total attempts.

**Layer 4: `guarded_run` doesn't retry output validation failures.**
It only catches `UnexpectedModelBehavior` containing `"RECITATION"` ([`src/aegis_ai/features/__init__.py:164-170`](src/aegis_ai/features/__init__.py)). All other `UnexpectedModelBehavior` exceptions propagate.

**Layer 5: Diagnostic data is lost.**
The `_run` handler at [line 114](src/aegis_ai/features/__init__.py) logs only `e.__class__.__name__` at WARNING and the full message at DEBUG. CI runs at INFO, so the actual validation error (which field failed, why) is invisible in the log. This makes post-mortem impossible without reproducing locally with DEBUG enabled.

### Evidence from the log

- **00:53:11** -- Classifier pre-computes IMPORTANT (confidence=0.86), `SuggestImpact(CVE-2023-53764) = ?` starts
- **00:53:43** -- `SuggestImpact(CVE-2023-53764) raised an exception: UnexpectedModelBehavior` (32 seconds, consistent with 2 LLM attempts)
- No "Using pre-computed classifier result" message for this CVE, indicating the LLM never called `kernel_impact_tool`
- This is the **only** `UnexpectedModelBehavior` in the entire eval run (all other kernel CVEs succeeded)

### Why this specific CVE

CVE-2023-53764 is unusual: it involves the wifi/mac80211 subsystem with `AV:A` (Adjacent Network, rare for kernel CVEs -- most use `AV:L`). The uncommon attack vector and networking context may push the model toward non-standard output patterns, but without DEBUG-level logs we cannot confirm which field failed validation.
</details>


Since we don't have debugging enabled in CI, we won't see the exception
https://github.com/RedHatProductSecurity/aegis-ai/blob/ab4767d1790d1c93b9402671389aebeae505a84f/src/aegis_ai/features/__init__.py#L112-L116

I'm running this locally `AEGIS_LLM_MAX_JOBS=1 uv run pytest -vv -s -o log_cli_level=DEBUG --show-capture=no evals`.

kdudka (approved):
Looks good.  Let's merge this once the CI is green.

PR #677, opened 2026-06-03
Remedy OOM issues

superbuggy:
## What
Bounds memory use in `KernelImpactClassifier` by rejecting oversized patch and commit HTML responses (via `Content-Length` and post-fetch size checks), capping commit hashes at 20, and adding `[mem]` checkpoint logging across the osidb bot pipeline (`util.log_memory`, `bot.py`, `suggest.py`, `kernel_cves` clone, and CLI entry).

## Why
The osidb bot was OOM-killing on CVEs with pathologically large kernel patches or GitHub commit pages (AEGIS-451); the limits skip unbounded buffers and the logging makes it easier to pinpoint which stage spikes RSS in production.

## How to Test
- Run the bot against a CVE known to trigger large kernel fetches: `aegis osidb-bot --cve-id <CVE>`
- Confirm logs show `[mem]` checkpoints at CLI entry, per-CVE start/end, feature execution, and classifier fetch stages
- For CVEs with many or oversized commits, verify warning logs for skipped patches/HTML and that the process completes without OOM
- Optionally run the existing test suite: `pytest`

## Related Tickets
https://redhat.atlassian.net/browse/AEGIS-451

### DIFF
--- a/src/aegis_ai/kernel_classifier/__init__.py
+++ b/src/aegis_ai/kernel_classifier/__init__.py
@@ -139,6 +139,8 @@ def _select_best_external_cvss3(cvss_scores: list[dict]) -> tuple[str, float, st
 
 
 _PATCH_SIZE_LIMIT = 1_000_000
+_HTML_SIZE_LIMIT = 5_000_000
+_MAX_COMMIT_HASHES = 20
 
 _BACKPORT_SIMILARITY_THRESHOLD = 0.75
 
@@ -400,7 +402,10 @@ class KernelImpactClassifier:
         """Fetch raw patches from git.kernel.org for given commit hashes.
 
         Tries torvalds, stable, linux-next, and the gregkh/linux GitHub
-        fallback in order.
+        fallback in order.  Responses exceeding ``_PATCH_SIZE_LIMIT`` are
+        rejected early via ``Content-Length`` or a streaming byte cap so
+        pathological commits never sit fully buffered in memory.
+
         Returns list of (commit_hash, patch_content) tuples.
         """
         patches = []
@@ -412,8 +417,31 @@ class KernelImpactClassifier:
                     url = tmpl.format(hash=commit_hash)
                     try:
                         resp = await client.get(url, follow_redirects=True)
-                        if resp.status_code == 200 and len(resp.text) > 100:
-                            patches.append((commit_hash, resp.text))
+                        if resp.status_code != 200:
+                            continue
+
+                        content_length = resp.headers.get("content-length")
+                        if content_length and int(content_length) > _PATCH_SIZE_LIMIT:
+                            logger.warning(
+                                "Skipping oversized patch %s from %s "
+                                "(%s bytes via Content-Length)",
+                                commit_hash[:12],
+                                tmpl.split("/")[2],
+                                content_length,
+                            )
+                            continue
+
+                        text = resp.text
+                        if len(text) > _PATCH_SIZE_LIMIT:
+                            logger.warning(
+                                "Dropping oversized patch %s (%d chars)",
+                                commit_hash[:12],
+                                len(text),
+                            )
+                            continue
+
+                        if len(text) > 100:
+                            patches.append((commit_hash, text))
                             logger.debug("Fetched patch for %s", commit_hash[:12])
                             fetched = True
                             used_template = tmpl
@@ -444,7 +472,8 @@ class KernelImpactClassifier:
         The al-kernel daemon analyses GitHub commit pages which contain crash
         reports, call stacks, and rendered diffs that raw git patches lack.
         We replicate that by fetching from GitHub (primary) with a
-        git.kernel.org fallback.
+        git.kernel.org fallback.  Responses exceeding ``_HTML_SIZE_LIMIT``
+        are rejected early to prevent large commits from bloating memory.
 
         Returns list of (commit_hash, html_content) tuples.
         """
@@ -456,8 +485,31 @@ class KernelImpactClassifier:
                     url = tmpl.format(hash=commit_hash)
                     try:
                         resp = await client.get(url, follow_redirects=True)
-                        if resp.status_code == 200 and len(resp.text) > 200:
-                            pages.append((commit_hash, resp.text))
+                        if resp.status_code != 200:
+                            continue
+
+                        content_length = resp.headers.get("content-length")
+                        if content_length and int(content_length) > _HTML_SIZE_LIMIT:
+                            logger.warning(
+                                "Skipping oversized commit HTML %s from %s "
+                                "(%s bytes via Content-Length)",
+                                commit_hash[:12],
+                                tmpl.split("/")[2],
+                                content_length,
+                            )
+                            continue
+
+                        text = resp.text
+                        if len(text) > _HTML_SIZE_LIMIT:
+                            logger.warning(
+                                "Dropping oversized commit HTML %s (%d chars)",
+                                commit_hash[:12],
+                                len(text),
+                            )
+                            continue
+
+                        if len(text) > 200:
+                            pages.append((commit_hash, text))
                             logger.debug("Fetched commit HTML for %s", commit_hash[:12])
                             fetched = True
                             break
@@ -576,6 +628,15 @@ class KernelImpactClassifier:
         if not commit_hashes:
             logger.info("No commit hashes for %s, skipping kernel classifier", cve_id)
             return None
+
+        if len(commit_hashes) > _MAX_COMMIT_HASHES:
+            logger.warning(
+                "Capping commit hashes for %s from %d to %d",
+                cve_id,
+                len(commit_hashes),
+                _MAX_COMMIT_HASHES,
+            )
+            commit_hashes = commit_hashes[:_MAX_COMMIT_HASHES]
 
         from aegis_ai.osidb_bot.util import log_memory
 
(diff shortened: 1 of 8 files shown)


kdudka:
@superbuggy I do not understand how any of these changes would prevent the captured OOM scenarios from occurring.  Were you able to reproduce the bug?

The `git clone` command itself has peak memory usage around 600M, which does not sounds like something that should trigger OOM.  There should be only one instance of `git clone` running at a time for a single `aegis osidb-bot` process.  These processes never run in parallel and definitely not in a single container.

As far as I can see, nothing in the captured logs indicates that the `git clone` command was running when the OOM state was reached.

superbuggy:
> @superbuggy I do not understand how any of these changes would prevent the captured OOM scenarios from occurring. Were you able to reproduce the bug?

I was not.

>  There should be only one instance of `git clone` running at a time for a single `aegis osidb-bot` process. These processes never run in parallel and definitely not in a single container.

It seemed plausible that overlapping containers would experience some memory pressure from the cloning operation in the situation where one contained process would be gathering patches for a CVE rich in patch data while another process began. However, I think this is not what is actually happening and moreover it shouldn't be possible, so this was the wrong approach.

This actually seems to happen when there is are several kernel CVEs being processed at once in a single process.

kdudka:
> This actually seems to happen when there is are several kernel CVEs being processed at once in a single process.

Yes, I tried to reproduce this in a debug pod in the stage infra (with a patched read-only osidb-bot) but the memory consumption (peak around 1 GiB) was way below the configured limits.  I ran `osidb-bot` manually passing it exactly the sets of CVEs that were being processed in the captured logs when the OOM state was reached.  The intersection of these three sets is CVE-2026-46077 but I was unable to find any reason why this specific CVE would cause OOM.

Does Aegis (in particular the kernel classifier code) read unbounded data from external sources into memory?

It is possible that the data read from external sources looked differently when it happened compared to when we were trying to reproduce this.

superbuggy:
> Does Aegis (in particular the kernel classifier code) read unbounded data from external sources into memory?

The git patches are technically unbounded, or rather have partial bounding but shouldn't be ever large enough to cause memory exhaustion. The likeliest cause seems to be memory not being reclaims after page cache writes, but it seems like a very peculiar set of worst-case circumstances.

I've offered some logging of memory usage as a potential source of insight if you think that valuable, though this issue does not seem to be presenting itself with any real frequency.

kdudka:
I ran the code from this branch in a debug pod on stage and it confirmed my observation (peak around 1 GiB):
```
[...]
2026-06-09 08:55:33 [INFO] [osidb-bot] [mem] SuggestImpact_end(CVE-2026-46077): VmPeak:  2516596 kB VmSize:      2451984 kB VmRSS:        473448 kB cgroup=992Mi (src/aegis_ai/osidb_bot/util.py:41)
2026-06-09 08:55:33 [INFO] [osidb-bot] [mem] cve_end(CVE-2026-46077): VmPeak:    2516596 kB VmSize:      2451984 kB VmRSS:        473448 kB cgroup=992Mi (src/aegis_ai/osidb_bot/util.py:41)
2026-06-09 08:55:33 [INFO] [osidb-bot] [mem] batch_end: VmPeak:  2516596 kB VmSize:      2451984 kB VmRSS:        473448 kB cgroup=992Mi (src/aegis_ai/osidb_bot/util.py:41)
```

kdudka:
@superbuggy There is a number of places in `src/aegis_ai/kernel_classifier` where unbounded content from external sources is read into memory:

### 1. Patch responses fully buffered before size check

https://github.com/RedHatProductSecurity/aegis-ai/blob/7b2d7501e1deac9e16000bd55eb26c35d86f44a6/src/aegis_ai/kernel_classifier/__init__.py#L414-L416

The entire HTTP response is materialized as `resp.text` before any size check. The `_filter_oversized` guard exists but only runs later inside `_deduplicate_patches` (called at [line 639](https://github.com/RedHatProductSecurity/aegis-ai/blob/7b2d7501e1deac9e16000bd55eb26c35d86f44a6/src/aegis_ai/kernel_classifier/__init__.py#L639)). Between fetch and filter, all patches sit in memory unbound. A single pathological response (e.g. the initial kernel commit at ~200 MB) would be fully buffered.

### 2. HTML commit pages fetched with no size limit at all

https://github.com/RedHatProductSecurity/aegis-ai/blob/7b2d7501e1deac9e16000bd55eb26c35d86f44a6/src/aegis_ai/kernel_classifier/__init__.py#L458-L460

HTML commit pages have **no** equivalent of `_PATCH_SIZE_LIMIT` / `_filter_oversized`. GitHub commit pages for large commits can be tens of MB. All pages for all commit hashes accumulate in the `pages` list without any cap.

### 3. No cap on number of commit hashes processed

https://github.com/RedHatProductSecurity/aegis-ai/blob/7b2d7501e1deac9e16000bd55eb26c35d86f44a6/src/aegis_ai/kernel_classifier/__init__.py#L407-L408

Both `_fetch_patches` and `_fetch_commit_html` iterate over `commit_hashes` without any limit on list length. Each commit triggers up to 4 HTTP fetches for patches and 2 for HTML, all accumulating in memory.

### Summary

| Location | What's unbounded | Mitigation present |
|---|---|---|
| [`_fetch_patches` L414](https://github.com/RedHatProductSecurity/aegis-ai/blob/7b2d7501e1deac9e16000bd55eb26c35d86f44a6/src/aegis_ai/kernel_classifier/__init__.py#L414) | Individual patch response size | `_filter_oversized` exists but runs **after** all patches are in memory |
| [`_fetch_commit_html` L459](https://github.com/RedHatProductSecurity/aegis-ai/blob/7b2d7501e1deac9e16000bd55eb26c35d86f44a6/src/aegis_ai/kernel_classifier/__init__.py#L459) | Individual HTML response size | **None** |
| [`_fetch_patches` L407](https://github.com/RedHatProductSecurity/aegis-ai/blob/7b2d7501e1deac9e16000bd55eb26c35d86f44a6/src/aegis_ai/kernel_classifier/__init__.py#L407), [`_fetch_commit_html` L452](https://github.com/RedHatProductSecurity/aegis-ai/blob/7b2d7501e1deac9e16000bd55eb26c35d86f44a6/src/aegis_ai/kernel_classifier/__init__.py#L452) | Number of commits fetched | **None** |

The most impactful fix would be to enforce size limits **at fetch time** (e.g. streaming with a byte cap, or checking `Content-Length` before reading the body) rather than after, and to cap the number of commit hashes processed.

src/aegis_ai/kernel_classifier/__init__.py
```diff
@@ -412,8 +417,31 @@ async def _fetch_patches(self, commit_hashes: list[str]) -> list[tuple[str, str]
                     url = tmpl.format(hash=commit_hash)
                     try:
                         resp = await client.get(url, follow_redirects=True)
-                        if resp.status_code == 200 and len(resp.text) > 100:
-                            patches.append((commit_hash, resp.text))
+                        if resp.status_code != 200:
+                            continue
+
+                        content_length = resp.headers.get("content-length")
```
kdudka: Does it make sense to check the `content-length` header after the whole document body is already read into memory?

src/aegis_ai/kernel_classifier/__init__.py
```diff
@@ -412,8 +417,31 @@ async def _fetch_patches(self, commit_hashes: list[str]) -> list[tuple[str, str]
...
+                        text = resp.text
+                        if len(text) > _PATCH_SIZE_LIMIT:
+                            logger.warning(
+                                "Dropping oversized patch %s (%d chars)",
+                                commit_hash[:12],
+                                len(text),
+                            )
+                            continue
+
+                        if len(text) > 100:
```
kdudka, line 478: Please insert a comment to explain why 100 is the correct threshold.

src/aegis_ai/kernel_classifier/__init__.py
```diff
@@ -456,8 +485,31 @@ async def _fetch_commit_html(
...
+                        text = resp.text
+                        if len(text) > _HTML_SIZE_LIMIT:
+                            logger.warning(
+                                "Dropping oversized commit HTML %s (%d chars)",
+                                commit_hash[:12],
+                                len(text),
+                            )
+                            continue
+
+                        if len(text) > 200:
```
kdudka, line 530: Please insert a comment to explain why 200 is the correct threshold.

src/aegis_ai/osidb_bot/util.py
```diff
@@ -16,3 +17,25 @@ def process(
 
 
 logger = _OsidbBotLogger(logging.getLogger(__name__), {})
+
+
+def log_memory(label: str) -> None:
```
kdudka, line 22: If we want to keep the memory usage logging hook, it should be disabled by default.  It seems overly disruptive to me for the regular use on production.

kdudka (changes requested):

src/aegis_ai/kernel_classifier/__init__.py
superbuggy: Changed to a stream check before body is buffered in.

src/aegis_ai/osidb_bot/util.py
kdudka: @superbuggy I have pushed commit 63319e67c9277da3df454403adc77d510527ac9d to ensure that the logging helper does not have any unintended side effects on production.

kdudka (approved):
@superbuggy Thanks for the update!  I have added one more commit to ensure that the logging helper does not have any unintended side effects on production.

PR #686, opened 2026-06-15
feat(osidb-bot): create kpanic labels from kernel classifier flags

superbuggy:
## What

  Add kernel classifier flag-to-label pipeline: when the kernel classifier detects
  `kernel_panic` or `kernel_panic_plus_uaf` active features, map them to a `kpanic`
  OSIDB alias label. The flags are stored as a `PrivateAttr` on `SuggestImpactModel`
  so they never appear in the LLM output schema or serialized model data. `FlawUpdater`
  creates the corresponding labels in OSIDB via `flaws.labels.create`, with graceful
  fallback on failure.

  ## Why

  Kernel panic CVEs need to be tagged with a `kpanic` label in OSIDB for
  downstream triage workflows. The classifier already identifies these features — this
  wires the signal through to label creation while keeping the flags invisible to the
  LLM (they are a classifier-only concern, not something the model should generate).

  ## How to Test

  ```bash
  uv run aegis osidb-bot --force CVE-2025-38718 CVE-2026-22998 CVE-2026-23011 CVE-2023-54222
  ```

  Verify:
  - _FLAG_TO_LABEL mapping tests pass
  - _flags is absent from SuggestImpactModel.model_json_schema() and model_dump()
  - suggest_impact() persists kernel_flags to aegis_meta when flags are present
  - FlawUpdater.do() calls labels.create with the correct payload
  - Label creation failure is logged but does not abort the flaw update

##  Related Tickets

Related: https://redhat.atlassian.net/browse/AEGIS-438

## Summary by Sourcery

Propagate kernel classifier flags into OSIDB by persisting them in flaw metadata and creating corresponding alias labels during flaw updates.

New Features:
- Map kernel classifier flags to OSIDB alias labels and persist them in flaw aegis_meta for provenance and downstream processing.

Enhancements:
- Extend SuggestImpactModel with a private flags attribute populated from classifier results while keeping it out of public schemas and serialized output.

Tests:
- Add comprehensive tests covering flag-to-label mapping, SuggestImpactModel flags behavior, aegis_meta.kernel_flags persistence, and label creation (including failure handling) in FlawUpdater.


src/aegis_ai/features/cve/__init__.py
```diff
@@ -506,6 +513,12 @@ async def exec(self, cve_id: CVEID, static_context: Any = None):
         result.output._classifier_diagnostics = classifier_result
         result.output._reconciliation_trace = trace
 
+        if classifier_result and isinstance(classifier_result, dict):
```
kdudka, line 520: Please insert a comment to explain why we store the list of flags.  The `_flags` field is undocumented in the data model, too.

src/aegis_ai/osidb_bot/suggest.py
```diff
@@ -207,6 +208,12 @@ async def suggest_impact(agent: Agent, flaw_data: FlawData, ts: datetime) -> set
     # record aegis_meta for RH CVSS (in the format used by OSIM)
     update_field(flaw_data, ts, "_cvss3_vector", output, value=output.cvss3_vector)
 
+    # persist kernel classifier flags for provenance and label creation
+    if output._flags:
+        aegis_meta = flaw_data.setdefault("aegis_meta", {})
+        aegis_meta[_KERNEL_FLAGS_KEY] = output._flags
```
kdudka: If the value is really meant to be written into the `aegis_meta` structure in OSIDB, it should use the same structure as all other fields.  This was initially implemented in `update_field()` but it is already being refactored in https://github.com/RedHatProductSecurity/aegis-ai/pull/683/changes/455d43a8caa379767f348b02de403f9d6a67aa11.

src/aegis_ai/osidb_bot/bot.py
```diff
@@ -252,6 +252,28 @@ async def do(self) -> None:
                     form_data=rh_cvss,
                 )
 
+            if _KERNEL_FLAGS_KEY in self.updated_fields:
```
kdudka: This needs to go to a separate method.

src/aegis_ai/osidb_bot/bot.py
```diff
@@ -252,6 +252,28 @@ async def do(self) -> None:
...
+                                "label": label_name,
+                                "type": "alias",
+                                "state": "NEW",
+                            },
+                        )
+                    except Exception as label_err:
+                        self._warn(
+                            f"failed to create label '{label_name}' "
+                            f"({label_err.__class__.__name__}: {label_err})"
+                        )
```
kdudka: If the label creation fails, the field needs to be removed from `self.updated_fields`, like we do for `cvss_scores`: https://github.com/RedHatProductSecurity/aegis-ai/blob/f43ad040e3d8e94c48a599bd6dfb0d49c5cc4d7a/src/aegis_ai/osidb_bot/bot.py#L259

kdudka: Also the full exception should not be logged in a non-debug mode.

kdudka (changes requested):
@superbuggy Please resolve Sourcery AI review comments before requesting additional reviews.  A few more comments inline...

kdudka (commented):
@superbuggy Thanks for the update!  Please squash fix-up commits and provide a list of CVEs for which the label is expected to be created in OSIDB so that we can easily test this in the stage environment.

superbuggy:
> @superbuggy Thanks for the update! Please squash fix-up commits and provide a list of CVEs for which the label is expected to be created in OSIDB so that we can easily test this in the stage environment.

Sure, I think we would need to manually reingest these flaws:

```
CVE-2025-38718 
CVE-2026-22998 
CVE-2026-23011
CVE-2023-54222 
```

kdudka:
@superbuggy Please optimize commit messages for reading.  You have squashed the fix-up commits, which is good.  But the resulting commit message (318fcc7bcdfb9f76d3c8003f2a2fffceefcafa21) is a concatenation of the original commit messages with all the redundant headers and footers.  Reading commit messages like this is really painful.  As you are using Claude to do commits, you can simply ask Claude to format a meaningful commit message for the actually committed changes.

src/aegis_ai/osidb_bot/suggest.py
```diff
@@ -248,6 +249,24 @@ async def suggest_impact(agent: Agent, flaw_data: FlawData, ts: datetime) -> set
     }
     flaw_data["cvss_scores"] = [rh_cvss]
     changed.add("cvss_scores")
+
+    # record aegis_meta for RH CVSS (in the format used by OSIM)
+    update_field(flaw_data, ts, "_cvss3_vector", output, value=output.cvss3_vector)
```
kdudka: It seems that the rebase went wrong.  This call should not be duplicated.  See commit c237c8cf4dbd89216fc9df57a24ce04ea7adbab2.

kdudka (changes requested):
Thanks for the update!  Please remove the duplicated call to `update_field()`.

kdudka:
@superbuggy I have updated the commit message per my [yesterday's comment](https://github.com/RedHatProductSecurity/aegis-ai/pull/686#issuecomment-4843400560), which you neither resolved nor replied on.

src/aegis_ai/osidb_bot/suggest.py
kdudka: Appears fixed in 522a6fa2af91cf0dc6a3ebe7cffbe923564987c2.

kdudka (approved):

PR #689, opened 2026-06-19
Feat/configure pre commit hooks

mpapadopoullos:
## What
adding pre-commit hooks

## Why
make check is manual

## How to Test

uvx pre-commit install

## Related Tickets
N/A

## Summary by Sourcery

Add and configure pre-commit hooks for code quality, formatting, and secret scanning.

New Features:
- Introduce a pre-commit configuration to run linting, formatting, secret scanning, and basic file hygiene checks on commit.

Enhancements:
- Add pre-commit to the development dependencies in pyproject.toml to support the new hook workflow.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,21 @@
+repos:
+  - repo: https://github.com/astral-sh/ruff-pre-commit
+    rev: v0.15.18
+    hooks:
+      - id: ruff
+        args: [--fix]
+      - id: ruff-format
+
+  - repo: https://github.com/gitleaks/gitleaks
+    rev: v8.30.0
+    hooks:
+      - id: gitleaks
+
+  - repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v6.0.0
+    hooks:
+      - id: trailing-whitespace
+      - id: end-of-file-fixer
+      - id: check-yaml
+      - id: check-toml
+      - id: check-merge-conflict
--- a/DEVELOP.md
+++ b/DEVELOP.md
@@ -94,6 +94,24 @@ We enforce code quality using **`ruff`** for linting and formatting. These check
     uvx ruff format .
     ```
 
+### Pre-commit Hooks
+
+The same linting and formatting checks can run automatically on every commit via [pre-commit](https://pre-commit.com/). To enable:
+
+```bash
+uv run pre-commit install
+```
+
+From that point on, each `git commit` will run ruff (lint + format), gitleaks (secret scanning), and a handful of file-hygiene checks (trailing whitespace, YAML/TOML validation, merge-conflict markers).
+
+> **Caution:** When a file-modifying hook fires, pre-commit will fail the commit and leave the fixed files as unstaged changes in the working tree. The developer then needs to review, `git add`, and commit again. So nothing lands in a commit without the developer seeing it, but **files on disk do get rewritten**.
+
+To run all hooks against the entire codebase (useful before opening a PR):
+
+```bash
+uv run pre-commit run --all-files
+```
+
 -----
 
 ## Configuration
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -105,6 +105,7 @@ allow-direct-references = true
 dev = [
     "build>=1.2.2",
     "detect-secrets>=1.5.0",
+    "pre-commit>=4.2.0",
     "exceptiongroup>=1.3.0",
     "hatchling>=1.27.0",
     "ipdb>=0.13.13",


kdudka (changes requested):
@mpapadopoullos Thanks for the contribution!  Please squash the fix-up commit per https://github.com/RedHatProductSecurity/aegis-ai/blob/0dc15ac296c89ddeae4b8a78e1c932715300704e/DEVELOP.md#pull-requests and document the pre-commit hook usage in `DEVELOP.md`.

At minimum we should say how to enable `pre-commit` and what to expect from this hook.  I would include also the following caution:
> When a file-modifying hook fires, pre-commit will fail the commit and leave the fixed files as unstaged changes in the working tree. The developer then needs to review, `git add`, and commit again. So nothing lands in a commit without the developer seeing it, but **files on disk do get rewritten**.

DEVELOP.md
```diff
@@ -94,6 +94,24 @@ We enforce code quality using **`ruff`** for linting and formatting. These check
...
+```
+
+From that point on, each `git commit` will run ruff (lint + format), gitleaks (secret scanning), and a handful of file-hygiene checks (trailing whitespace, YAML/TOML validation, merge-conflict markers).
+
+> **Caution:** When a file-modifying hook fires, pre-commit will fail the commit and leave the fixed files as unstaged changes in the working tree. The developer then needs to review, `git add`, and commit again. So nothing lands in a commit without the developer seeing it, but **files on disk do get rewritten**.
+
+To run all hooks against the entire codebase (useful before opening a PR):
+
+```bash
+uv run pre-commit run --all-files
```
kdudka, line 112: @mpapadopoullos Did you actually try to run this command in the Aegis git repo?

It causes a disaster by overwriting all evals fixtures in `evals/*_cache` and data files in `src/aegis_ai_ml/src/classifier/kernel-cve-impact-classifier/data`.

mpapadopoullos: no.
1. by default, the pre-commit hooks will run on staged files (when installed)
2. we can add exception to evals/ directory, so they are ignored

example:
```
diff --git a/.pre-commit-config.yaml b/.pre-commit-config.yaml
index 4e8bc1c..e5e5612 100644
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,3 +1,5 @@
+exclude: '(evals)/'
+
 repos:
   - repo: https://github.com/astral-sh/ruff-pre-commit
     rev: v0.15.18
```

ref: https://pre-commit.com/#top_level-exclude

kdudka: That does not seem correct either.  We do have source code files in `/evals` which should be checked while changing them in a commit.  Yet it does not exclude files not under `/evals` which should be excluded.  We should sync the pre-commit hooks with the related CI checks to avoid unnecessary surprises.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,21 @@
...
+      - id: gitleaks
+
+  - repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v6.0.0
+    hooks:
+      - id: trailing-whitespace
+      - id: end-of-file-fixer
+      - id: check-yaml
+      - id: check-toml
+      - id: check-merge-conflict
```
superbuggy, line 21: I would prefer to use at least the same commands as the make targets so we don't have to deal with dependency drift and maintenance for the pre commit hooks,  however our make targets don't parameterize for filenames. This introduces some conflict with conventions/intended usage around the usage of pre commit hooks. That might look something like this:

```yaml  repos:
    - repo: local
      hooks:
        - id: lint
          name: ruff check
          entry: make lint
          language: system
          pass_filenames: false
          types: [python]

        - id: format
          name: ruff format
          entry: make format
          language: system
          pass_filenames: false
          types: [python]

# probably this too slow for a pre commit hook, but in the PR comment you reference make check which runs this:
        - id: ty-check
          name: ty check
          entry: make check-type
          language: system
          pass_filenames: false
          types: [python]

    - repo: https://github.com/gitleaks/gitleaks
      rev: v8.30.0
      hooks:
        - id: gitleaks
```

mpapadopoullos: makes sense. if you are OK with the idea of pre-commit hooks and this configuration I can push the necessary changes.

kdudka: Yes, please.  Update the PR and we will try it out.

mpapadopoullos: Perfect. Will do this tmr.

kdudka:
This pull request has not been updated since June.  @mpapadopoullos If you still want to contribute this feature to Aegis, please open a fresh PR (reflecting the review comments provided on this PR).

PR #711, opened 2026-07-02
feat: add KPI metrics for osidb-bot auto-processed flaws

superbuggy:
## What

Adds `GET /api/v1/analysis/kpi/osidb-bot`, which queries OSIDB for bot-processed flaws (`workflow_state=DONE`) and compares osidb-bot's suggestions against the current flaw values. Reports per-field metrics (applied, skipped, kept, modified, acceptance rate, avg data quality/confidence, avg distance) for `components`, `title`, `cve_description`, `cwe_id`, `impact`, and the CVSS3 vector, with an incremental file-based cache and optional `changed_after`/`changed_before`/`full_refresh` query params.

Follow-up commits fix issues found in code review:
- Compare against the latest bot suggestion instead of the oldest, and dedupe flaws re-seen across incremental fetches
- Prevent cache corruption when `full_refresh` is combined with date filters
- Handle `cvss_scores: null` and non-numeric `data_quality`/`confidence` values without crashing the whole computation
- Serialize the cache's read-modify-write cycle with a file lock so concurrent requests can't clobber each other's results
- Stamp the incremental-fetch cutoff before the OSIDB query runs (not after), so flaws updated mid-fetch aren't permanently skipped
- Narrow the OSIDB-connection error handler to `OSError` so real bugs surface as 500s instead of a misleading 503
- Consolidate the duplicated bot-processed-flaw filter/traversal in `aggregate_kpi` and `_extract_flaw_ids`

## Why

Product Security wants visibility into how often analysts keep vs. override osidb-bot's automated suggestions, to gauge the bot's accuracy per field. The follow-up fixes close correctness and concurrency gaps from code review that could otherwise silently drop or corrupt KPI data in production.

## How to Test

- `uv run pytest tests/test_bot_kpi_cache.py src/aegis_ai_web/tests/test_bot_kpi.py src/aegis_ai_web/tests/test_bot_kpi_endpoint.py` (90 tests)
- `make test test-web` for the full suite
- Manual: `curl /api/v1/analysis/kpi/osidb-bot` against a live/staging OSIDB, then repeat immediately to confirm the incremental cache path is used; try `full_refresh=true` and `changed_after`/`changed_before` filters

## Related Tickets

https://redhat.atlassian.net/browse/AEGIS-445

---

## Summary by Sourcery

Expose reliable osidb-bot KPI reporting with incremental caching and per-field acceptance metrics.

New Features:
- Add an API endpoint for per-field osidb-bot KPI metrics covering suggestions, skips, acceptance, quality, confidence, and deviation.
- Support optional date-range filtering and incremental, file-backed caching for KPI queries.

Bug Fixes:
- Ensure KPI results remain correct across repeated suggestions, changing flaws, malformed OSIDB data, cache corruption, concurrent requests, and incremental fetch boundaries.
- Return distinct service-unavailable and internal-error responses for OSIDB failures.

Enhancements:
- Extract reusable impact and CVSS comparison utilities and a generic locked JSON state-file handler.
- Refine bot KPI aggregation to process the latest suggestion and reconcile cached flaw state accurately.

Documentation:
- Document the KPI cache directory setting and expose the new endpoint and response schemas in the API specification.

Tests:
- Add comprehensive unit, endpoint, cache, filtering, incremental-fetch, and concurrency coverage for osidb-bot KPI reporting.


src/aegis_ai/osidb_bot/kpi.py
```diff
@@ -0,0 +1,225 @@
+"""KPI metrics extraction for osidb-bot auto-processed flaws.
```
kdudka: This does not belong to `aegis_ai/osidb_bot`.  `aegis-web` and `osidb-bot` are intentionally completely isolated from each other.  They use separate volumes to store persistent files and different ways to access OSIDB.  `osidb-bot` already records the needed metadata to the `aegis_meta` namespace in OSIDB.  The KPI endpoint needs to take it from there and possibly cache aggregated metrics server-side independently on `osidb-bot` (which could be tricky because the subset of visible OSIDB data may depend on identity of the user requesting the statistics over the KPI endpoint).

src/aegis_ai/osidb_bot/kpi.py
```diff
@@ -0,0 +1,225 @@
...
+from pydantic import BaseModel
+
+from osidb_bindings.session import Session
+
+logger = logging.getLogger(__name__)
+
+# Flaw fields that the bot writes suggestions for and that we can compare
+# against current values.  Underscore-prefixed keys (_ecosystems, _cvss3_vector)
+# are metadata-only and have no direct flaw-field counterpart.
+BOT_KPI_FIELDS = ("components", "title", "cve_description", "cwe_id", "impact")
```
kdudka: RH CVSS is missing in the tuple.

src/aegis_ai/osidb_bot/kpi.py
```diff
@@ -0,0 +1,225 @@
...
+        return round(self.confidence_sum / self.total_entries, 2)
+
+
+@dataclass
+class BotKPIResult:
+    total_flaws_processed: int = 0
+    features: dict[str, FeatureStats] = field(default_factory=dict)
+
+
+def _values_equal(suggested: Any, current: Any) -> bool:
```
kdudka: Do we only want to check the values for equality?  I would use distance at least for impact and CVSS, like we do in evals.  Arguably, when Aegis misclassifies IMPORTANT impact as CRITICAL, it is less of an issue compared to when it misclassifies IMPORTANT as NONE.

kdudka (changes requested):

src/aegis_ai_web/src/endpoints/bot_kpi.py
```diff
@@ -0,0 +1,533 @@
...
+        else:
+            result = _fetch_with_cache(osidb)
+        return _result_to_response(result)
+    except OSError:
+        # OSError also covers requests.exceptions.RequestException (it subclasses
+        # OSError), so this catches real connectivity failures without masking
+        # programming bugs (AttributeError, TypeError, etc.) as a 503 outage.
+        logger.error("Failed to connect to OSIDB")
+        logger.debug("OSIDB connection error details", exc_info=True)
+        raise HTTPException(
```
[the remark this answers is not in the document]
superbuggy: @sourcery-ai `RequestException` subclasses `OSError`, this is erroneous.

src/aegis_ai_web/src/main.py
```diff
@@ -761,6 +764,75 @@ async def cve_kpi(
...
+        description="Force a full query from scratch, ignoring and overwriting the cache.",
+    ),
+) -> BotKPIResponse:
+    """Get KPI metrics for flaws auto-processed by the osidb-bot."""
+    return await asyncio.to_thread(
+        get_osidb_bot_kpi,
+        changed_after=changed_after,
+        changed_before=changed_before,
+        full_refresh=full_refresh,
+    )
```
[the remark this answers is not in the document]
superbuggy: Spurious, this is handled in `main.py` and not via endpoint source code modules.

src/aegis_ai_web/src/data_models.py
```diff
@@ -331,6 +331,67 @@ def validate_csv_headers(csv_headers: list[str]) -> bool:
...
+    )
+    skipped: int = Field(
+        ...,
+        ge=0,
+        description="Suggestions discarded by quality gates (type=AI-Bot-Skipped)",
+    )
+    kept: int = Field(
+        ...,
+        ge=0,
+        description="Applied suggestions still unchanged in OSIDB after analyst review",
```
kdudka: Please remove ` after analyst review` from the description.  We do not know what changed the value.  It is more and more likely that the value was changed by some automation rather than an analyst these days.

src/aegis_ai_web/src/data_models.py
```diff
@@ -331,6 +331,67 @@ def validate_csv_headers(csv_headers: list[str]) -> bool:
...
+    )
+    modified: int = Field(
+        ...,
+        ge=0,
+        description="Applied suggestions changed by an analyst",
```
kdudka: Same here.  Remove ` by an analyst`.

src/aegis_ai_web/src/data_models.py
```diff
@@ -331,6 +331,67 @@ def validate_csv_headers(csv_headers: list[str]) -> bool:
...
+    )
+
+
+class BotKPIResponse(BaseModel):
+    """Response model for osidb-bot KPI metrics endpoint."""
+
+    total_flaws_processed: int = Field(
+        ...,
+        ge=0,
+        description="Total flaws processed by the bot and reviewed by analysts",
```
kdudka: Similar concern here about `reviewed by analysts`.  How would we know which flaws were reviewed by an analyst with the automated workflows?

src/aegis_ai_web/src/main.py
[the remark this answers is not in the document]
superbuggy: Your finding describes a pre-existing, application-wide design that is not related to this PR. Authentication toggling is a deployment-time concern, as AEGIS_WEB_SPN is set in the environment. it is spurious since this PR does not introduce this issue and it implicitly it is out of scope.

src/aegis_ai_web/src/endpoints/bot_kpi.py
```diff
@@ -0,0 +1,547 @@
...
+def _score_bot_entries(
+    entries: list[dict[str, Any]],
+) -> tuple[FeatureStats, dict[str, Any] | None]:
+    """Accumulate per-entry stats and return the latest AI-Bot entry, if any."""
+    stats = FeatureStats()
+    bot_entry: dict[str, Any] | None = None
+
+    for entry in entries:
+        stats.data_quality_sum += _safe_float(entry.get("data_quality", 0))
+        stats.confidence_sum += _safe_float(entry.get("confidence", 0))
```
kdudka: Missing `data_quality` or `confidence` values should be treated as unknown rather than 0.0.  The recording of these metrics was introduced later: 8702cd69cff0e07b14d939a01583acf98aee4076

src/aegis_ai_web/src/endpoints/bot_kpi.py
```diff
@@ -0,0 +1,547 @@
...
+    return flaws
+
+
+class BotKPICacheEntry(BaseModel):
+    """On-disk cache for precomputed bot KPI aggregates."""
+
+    cutoff: datetime
+    total_flaws_processed: int
+    features: dict[str, dict[str, float]]
+    flaw_ids: list[str] = []
```
kdudka: What about expiration of the cached data?  Do we expect that a CVE processed by Aegis today would get the same scores tomorrow (after the suggested values get actually reviewed)?

kdudka (changes requested):
@superbuggy Please resolve outstanding review comments before requesting an additional review.  It is exhausting to go through them repeatedly and check which of them have been addressed when they are still open with no reply.

src/aegis_ai/osidb_bot/kpi.py
superbuggy: Changed to use the same logic as the eval suite.

src/aegis_ai_web/src/endpoints/bot_kpi.py
superbuggy: these are now treated as unknown.

src/aegis_ai_web/src/endpoints/bot_kpi.py
superbuggy: I'm not quite sure, I updated it to rescore, but this may not be desirable in many cases, and it may be better to ~have a sliding window instead~ have  a max age.

superbuggy: Sorry for the misleading comment/noise. I should restate, a new request should trigger a cache miss when the updated_dt of a CVE changes, which seems fine.

src/aegis_ai_web/src/endpoints/bot_kpi.py
```diff
@@ -0,0 +1,588 @@
...
+    osidb: Any,
+    *,
+    changed_after: datetime | None = None,
+    changed_before: datetime | None = None,
+) -> list[dict[str, Any]]:
+    """Query OSIDB for bot-processed flaws in DONE workflow state."""
+    kwargs: dict[str, Any] = {
+        "include_fields": _INCLUDE_FIELDS,
+        "workflow_state": "DONE",
+        "aegis_meta__has_key": "processed",
```
kdudka: There is no `aegis_meta__has_key` predicate in OSIDB.  It seems to be hallucinated by a code assistant.
I do not believe this code has ever (successfully) run.

kdudka: Fixed in 8296aea32a50ea386d90e5151d4a8c1ed1835aac.

src/aegis_ai_web/src/main.py
```diff
@@ -761,6 +764,80 @@ async def cve_kpi(
...
+async def osidb_bot_kpi(
+    changed_after: datetime | None = Query(  # noqa: B008
+        default=None,
+        description="Only include flaws updated after this ISO 8601 datetime (e.g. 2025-01-01T00:00:00Z).",
+    ),
+    changed_before: datetime | None = Query(  # noqa: B008
+        default=None,
+        description="Only include flaws updated before this ISO 8601 datetime.",
+    ),
+    full_refresh: bool = Query(
```
kdudka: What is the use case for the `full_refresh` flag?

Under which circumstances would the client know better that the server-side cache needs to be refreshed?

src/aegis_ai_web/src/endpoints/bot_kpi.py
```diff
@@ -0,0 +1,691 @@
...
+    settings = get_settings()
+    if settings.bot_kpi_cache_dir:
+        return Path(settings.bot_kpi_cache_dir) / "kpi_cache.json"
+    return Path(settings.config_dir) / "osidb_bot_kpi" / "kpi_cache.json"
+
+
+@contextmanager
+def _cache_lock():
+    """Serialize cache read-modify-write cycles across concurrent requests."""
+    lock_path = _get_cache_path().with_suffix(".lock")
```
kdudka: Why do we need a separate file with `.lock` suffix?

It makes the state unnecessarily complex.  One needs to separately test the scenarios where one of the files exists while the other file does not exist.  I propose to rather lock the cache file itself.

See how this is implemented in `StateFileHandler` in `osidb_bot`:
https://github.com/RedHatProductSecurity/aegis-ai/blob/ad25643483e9967d4b2fa5440e26ddea9f14327f/src/aegis_ai/osidb_bot/state.py#L35

I believe this class can be easily generalized to handle arbitrary data model.

src/aegis_ai_web/src/endpoints/bot_kpi.py
```diff
@@ -0,0 +1,691 @@
...
+    edited again after the bot's initial pass: its entry is simply
+    overwritten and the aggregate is resummed from scratch, instead of the
+    flaw being permanently frozen at its first-seen state.
+    """
+
+    cutoff: datetime
+    # No defaults: a cache file from before this schema (flat aggregate
+    # sums + a seen-IDs list, no `flaws`/`aggregate` keys) must fail
+    # validation here so `_load_cache` treats it as absent and triggers a
+    # full refresh, rather than silently defaulting to an empty cache.
```
kdudka: I do not understand this comment.  `a cache file from before this schema` is irrelevant as long as there is no released version Aegis with this feature.

Also it is unclear to me how what the difference between `full refresh` and `silently defaulting to an empty cache` is.  Aegis should simply load and cache data from OSIDB as needed.

src/aegis_ai_web/src/endpoints/bot_kpi.py
```diff
@@ -0,0 +1,691 @@
...
+    flaws: dict[str, FlawCacheData]
+    aggregate: dict[str, dict[str, float]]
```
kdudka: As I understand it, aggregated values should **not** be cached because, they depend on the changes in OSIDB in between the queries, as well as on the query parameters.

There is no value in caching them.  The cache should help to avoid repeatedly fetching of unchanged data from OSIDB.  When you have the needed per-CVE data in memory, computing of the aggregated values is cheap.  The (de)serialization of the JSON cache is much more expensive than any float arithmetic operations anyway.

kdudka (changes requested):

src/aegis_ai_web/src/main.py
superbuggy: Removed `full_refresh` entirely (endpoint param, handler arg, and regenerated OpenAPI docs). There is no client-side use case: a flaw is now re-scored automatically whenever it changes in OSIDB, since every fetched flaw overwrites its own per-CVE cache entry and the aggregate is resummed from scratch each request. The client never needs to tell the server its cache is stale.

src/aegis_ai_web/src/endpoints/bot_kpi.py
superbuggy: Done. Generalized the bot's `StateFileHandler` into `aegis_ai/state_file.py` (`StateFileHandler[T: BaseModel]`), which locks the state file itself with `fcntl.flock` — no `.lock` sidecar. `osidb_bot/state.py` now subclasses it; the KPI endpoint uses it directly (blocking lock to serialize concurrent read-modify-write). The removed `.lock` file eliminates the two-file scenarios you noted.

src/aegis_ai_web/src/endpoints/bot_kpi.py
superbuggy: Removed the comment and the distinction it described. With `full_refresh` gone there is no "full refresh vs. empty cache" behavior — Aegis just loads and caches per-CVE data from OSIDB as needed and incrementally updates it. A missing/empty/corrupt cache degrades to "no cache" and is repopulated on the next fetch.

src/aegis_ai_web/src/endpoints/bot_kpi.py
superbuggy: Agreed — removed the cached `aggregate`. `BotKPICacheEntry` now stores only per-CVE data (`cutoff` + `flaws`). The aggregate is recomputed from the cached per-flaw stats on every request via `_resum()`, so it always reflects current OSIDB data and the request's date filters. The cache now exists solely to avoid re-fetching unchanged flaws from OSIDB.

src/aegis_ai/osidb_bot/state.py
```diff
@@ -1,13 +1,11 @@
...
 from datetime import datetime
-from typing import Any, Never, Self
+from typing import Any
 
-from pydantic import BaseModel, ConfigDict, Field, ValidationError
+from pydantic import BaseModel, ConfigDict, Field
 
 from aegis_ai.data_models import CVEID
 from aegis_ai.osidb_bot.util import logger
+from aegis_ai.state_file import StateFileHandler as _StateFileHandler
```
kdudka: This is confusing.  `StateFileHandler` denotes the base class in some context but a subclass in another context.  Please rename the subclass to `BotStateFileHandler` instead.

src/aegis_ai/state_file.py
```diff
@@ -0,0 +1,108 @@
...
+        state_file: str | None,
+        model_type: type[T],
+        *,
+        blocking: bool = False,
+        exclude_defaults: bool = False,
+    ):
+        self.state_file = state_file
+        self.model_type = model_type
+        self.blocking = blocking
+        self.exclude_defaults = exclude_defaults
```
kdudka: Why do we need to use different `exclude_defaults`  value for osidb-bot and KPI cache?  Is there any significant benefit in (de)serializing the default values on read/write?  If not, I would keep this simple and use `exclude_defaults=True` for both.

src/aegis_ai/state_file.py
```diff
@@ -0,0 +1,108 @@
...
+
+import fcntl
+import json
+import logging
+import os
+from typing import Never, Self
+
+from pydantic import BaseModel, ValidationError
+
+logger = logging.getLogger(__name__)
```
kdudka: I am afraid this unexpectedly changes the logging output for `osidb-bot`, which would otherwise use the `[osidb-bot] ` prefix: https://github.com/RedHatProductSecurity/aegis-ai/blob/cace9b746e1cb147da99550656ef04470bae2200/src/aegis_ai/osidb_bot/util.py#L10-L16

src/aegis_ai/state_file.py
```diff
@@ -0,0 +1,108 @@
...
+        if not self.state_file:
+            # no state file configured -> no-op handler
+            return self
+
+        try:
+            self.state_fd = os.open(self.state_file, os.O_RDWR | os.O_CREAT, 0o644)
+        except OSError as e:
+            self._fail(e, "failed to open or create state file")
+
+        flags = fcntl.LOCK_EX if self.blocking else fcntl.LOCK_EX | fcntl.LOCK_NB
```
kdudka, line 78: The use of a blocking syscall could be problematic in an asyncio-based application.  In this specific case, the code is reached through `asyncio.to_thread()`, which mitigates blocking of the main event loop.  But, at minimum, there should be a big warning about this in the docstring of `StateFileHandler` as this can lead to difficult to debug freezes in production when someone uses this (now generic) class in a different context.

src/aegis_ai_web/src/endpoints/bot_kpi.py
```diff
@@ -0,0 +1,613 @@
...
+# Flaw fields that the bot writes suggestions for and that we can compare
+# against current values.  Underscore-prefixed keys (_ecosystems, _cvss3_vector)
+# are metadata-only and have no direct flaw-field counterpart.
+BOT_KPI_FIELDS = ("components", "title", "cve_description", "cwe_id", "impact")
+
+_META_ONLY_FIELDS = ("_cvss3_vector",)
+_ALL_KPI_FIELDS = BOT_KPI_FIELDS + _META_ONLY_FIELDS
+
+_INCLUDE_FIELDS = ",".join(
+    ["cve_id", "updated_dt", "aegis_meta", "cvss_scores", *BOT_KPI_FIELDS]
```
kdudka, line 40: As discussed on Slack, we should only fetch `cve_id` and `updated_dt` in the initial search.  The other fields should be fetched per flaw only when the corresponding cache entry is missing or outdated.  Otherwise I do not understand how the cache would optimize the amount of data read from OSIDB.

src/aegis_ai_web/src/endpoints/bot_kpi.py
```diff
@@ -0,0 +1,613 @@
...
+    than the JSON (de)serialization the cache already pays for. The cache
+    exists solely to avoid re-fetching unchanged flaws from OSIDB.
+
+    Storing each flaw's contribution individually (rather than only a running
+    total) is what lets a flaw be correctly re-scored if it's edited again
+    after the bot's initial pass: its entry is simply overwritten and the
+    aggregate is resummed from scratch.
+    """
+
+    cutoff: datetime
```
kdudka: So you removed the comment that I did not understand.  But I still do not understand the field itself.  We should maintain per-CVE cache entries because there is no globally valid `cutoff` across API calls.

kdudka (changes requested):

src/aegis_ai/osidb_bot/state.py
superbuggy: Good point

src/aegis_ai/state_file.py
superbuggy: Dropped the per-call parameter — both osidb-bot and the KPI cache now serialize with `exclude_defaults=True` (c450e99c). There was no benefit to persisting default values, so keeping it uniform is simpler.

src/aegis_ai/state_file.py
superbuggy: Fixed in c450e99c. `StateFileHandler` now takes an injected `logger`, and `BotStateFileHandler` passes the bot's `[osidb-bot]` logger, so osidb-bot log lines keep their prefix.

src/aegis_ai/state_file.py
superbuggy: Added a `.. warning::` about the blocking `fcntl.flock` syscall to both the module and class docstrings in `state_file.py` (c450e99c). Confirmed the KPI endpoint reaches the blocking handler via `asyncio.to_thread()`.

src/aegis_ai/osidb_bot/state.py
superbuggy: Renamed the subclass to `BotStateFileHandler` in c450e99c. The shared generic base in `aegis_ai/state_file.py` keeps the `StateFileHandler` name, so the base class and the bot-specific specialization are no longer both called the same thing.

src/aegis_ai_web/src/endpoints/bot_kpi.py
superbuggy: Done in 39ac3033. `_fetch_flaw_index()` now fetches only `cve_id,updated_dt` for every DONE flaw; `_fetch_one_flaw()` fetches the full fields via `retrieve(id=...)` only when a flaw is new or its `updated_dt` has advanced past the cached one. Since bot-processed status cannot be filtered server-side, non-bot-processed DONE flaws are cached as skip markers (with no field stats) so their full data is not re-fetched until their watermark changes.

src/aegis_ai_web/src/endpoints/bot_kpi.py
superbuggy: Agreed. Removed the global `cutoff` field entirely (39ac3033). Each `FlawCacheData` is now keyed on its own `updated_dt` watermark, and the cache is rebuilt to exactly the current DONE index on every request (pruning flaws that have left DONE). A flaw is re-fetched only when its own timestamp advances — there is no longer any cross-request cutoff.

src/aegis_ai_web/src/endpoints/bot_kpi.py
```diff
@@ -0,0 +1,656 @@
...
+    """
+    kwargs: dict[str, Any] = {
+        "include_fields": "cve_id,updated_dt",
+        "workflow_state": "DONE",
+        "limit": 200,
+    }
+    logger.info("querying OSIDB flaw index: %s", kwargs)
+
+    index: dict[str, str] = {}
+    for flaw in osidb.flaws.retrieve_list_iterator(**kwargs):
```
kdudka, line 418: `kwargs` are missing the search predicates based on query parameters, such as `changed_after`, which is wrong.  I did a smoke test by querying KPI metrics for the last 3 days and after 10 minutes, the query was still running.  Based on the server-side log, Aegis was scraping the full OSIDB history:
```
2026-08-27 16:04:02 [INFO] querying OSIDB flaw index: {'include_fields': 'cve_id,updated_dt', 'workflow_state': 'DONE', 'limit': 200} (src/aegis_ai_web/src/endpoints/bot_kpi.py:308)
```

Even when the query has no parameters, there should be some reasonable bound to limit the search space, e.g. `created_dt_gte=osidb_bot_birthday`.  Otherwise Aegis would be overloading OSIDB for no good reason.

kdudka (changes requested):

src/aegis_ai_web/src/endpoints/bot_kpi.py
superbuggy: Fixed in 45b74c91. The request's `changed_after`/`changed_before` are now pushed down to OSIDB as `updated_dt_gte`/`updated_dt_lte`, so a narrow query fetches only that window instead of the whole history. When no `changed_after` is given the lower bound defaults to the osidb-bot "birthday" (`2026-02-10`, the feature's first commit).

I bounded on `updated_dt` rather than your `created_dt_gte` example: every bot-processed flaw was necessarily *updated* by the bot's own write at/after the birthday, so `updated_dt_gte=birthday` can never exclude one, whereas `created_dt_gte` would drop older CVEs the bot processed later.

src/aegis_ai_web/src/endpoints/bot_kpi.py
```diff
@@ -0,0 +1,684 @@
...
+    """
+    kwargs: dict[str, Any] = {
+        "include_fields": "cve_id,updated_dt",
+        "workflow_state": "DONE",
+        "updated_dt_gte": changed_after or OSIDB_BOT_BIRTHDAY,
+        "limit": 200,
+    }
+    if changed_before is not None:
+        kwargs["updated_dt_lte"] = changed_before
+    logger.info("querying OSIDB flaw index: %s", kwargs)
```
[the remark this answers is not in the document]
kdudka: There is no sensitive information in `kwargs` here.

src/aegis_ai_web/src/endpoints/bot_kpi.py
```diff
@@ -0,0 +1,684 @@
...
+    The search is always bounded by ``updated_dt``: the request's
+    ``changed_after``/``changed_before`` filters are pushed down to OSIDB so a
+    narrow query (e.g. the last few days) fetches only that window instead of
+    scraping the bot's entire history. When no ``changed_after`` is given the
+    lower bound defaults to :data:`OSIDB_BOT_BIRTHDAY`.
+    """
+    kwargs: dict[str, Any] = {
+        "include_fields": "cve_id,updated_dt",
+        "workflow_state": "DONE",
+        "updated_dt_gte": changed_after or OSIDB_BOT_BIRTHDAY,
```
kdudka: `OSIDB_BOT_BIRTHDAY` should be used unconditionally.  There is no practical use case to set `changed_after` lower than `OSIDB_BOT_BIRTHDAY`.  It would overload OSIDB for no real benefit.

src/aegis_ai_web/src/endpoints/bot_kpi.py
```diff
@@ -0,0 +1,684 @@
...
+    }
+    if changed_before is not None:
+        kwargs["updated_dt_lte"] = changed_before
+    logger.info("querying OSIDB flaw index: %s", kwargs)
+
+    index: dict[str, str] = {}
+    for flaw in osidb.flaws.retrieve_list_iterator(**kwargs):
+        flaw_dict = flaw.to_dict()
+        cve_id = flaw_dict.get("cve_id")
+        if not cve_id:
```
kdudka, line 421: Fetching flaws without CVE ID just to throw them away later on is wasteful.  Use the `"cve_id__isempty": False` search predicate instead.

src/aegis_ai_web/src/endpoints/bot_kpi.py
```diff
@@ -0,0 +1,684 @@
...
+        osidb, changed_after=changed_after, changed_before=changed_before
+    )
+
+    cached = _read_cache()
+    cached_flaws = cached.flaws if cached is not None else {}
+
+    fetched: dict[str, FlawCacheData] = {}
+    for cve_id, updated_dt in index.items():
+        if _needs_fetch(cached_flaws.get(cve_id), updated_dt):
+            flaw_data = _fetch_one_flaw(osidb, cve_id)
```
kdudka: @superbuggy Were you able to successfully test the KPI endpoint?

I again tried a toy query for the **last 7 days only** and gave up waiting after 10 minutes.  Fetching flaw data separately for each CVE takes approx 1s.  We have approx. 2000 CVEs in the DONE state that were updated in the last 7 days.  This means approx. 30min of waiting just to get last 7 days data.

See commit f86699668e9f2874a0f7a1160c343cc65a60dc2e, which implements batched retrieval.

kdudka (changes requested):

src/aegis_ai_web/src/endpoints/bot_kpi.py
kdudka: Actually I see that this is `updated_dt_gte` rather than `created_dt_gte` as proposed at https://github.com/RedHatProductSecurity/aegis-ai/pull/711#discussion_r3872584968.  You can simply use `created_dt_gte` for `OSIDB_BOT_BIRTHDAY` and keep `updated_dt_gte` for `changed_after`.  This will ensure no one would query unrelated OSIDB data, either by mistake, or for a DoS attack attempt.

superbuggy: Done in 63503f12. `OSIDB_BOT_BIRTHDAY` is now applied unconditionally as `created_dt_gte`, and `changed_after` is kept as `updated_dt_gte`. This guarantees the index search can never reach OSIDB data created before the bot existed, closing the mistake/DoS window you noted.

src/aegis_ai_web/src/endpoints/bot_kpi.py
superbuggy: Done in 63503f12. Added `cve_id__isempty=False` to the index query so CVE-less flaws are filtered server-side instead of fetched and discarded.

src/aegis_ai_web/src/endpoints/bot_kpi.py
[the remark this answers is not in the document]
superbuggy: Agreed, no sensitive data in `kwargs` (just the search predicates and date bounds). The CodeQL alert has been dismissed as a false positive.

src/aegis_ai_web/src/endpoints/bot_kpi.py
superbuggy: Thanks, I pulled your batched-retrieval commit (f866996) into the branch.

kdudka (changes requested):
@superbuggy I think we need to improve the handling of `changed_after` and `changed_before`.  I was really shocked that `acceptance_rate` for `components` in the last 7 days was only 71.5%.  So I investigated where the number came from.  The cause was that flaws, such as CVE-2026-40224, processed by Aegis/osidb-bot back in April (when components suggestion was still experimental) were updated the last week (new affects were added without touching the `components` field).  This is really misleading.

I think we should exclude ~~CVEs~~ _suggested fields_ where the **Aegis/osidb-bot suggestion timestamp** is not within the time frame specified in the query parameters.

superbuggy:
@kdudka Fixed in ae8a3629. Date filtering now keys on each suggestion's own recorded `timestamp` in `aegis_meta` rather than the flaw's `updated_dt`, so a suggested field is counted only when *its* suggestion falls within `changed_after`/`changed_before`. A flaw like CVE-2026-40224 — whose `components` suggestion was made in April but whose `updated_dt` moved last week because new affects were added — no longer contributes its stale `components` suggestion to a last-7-days query. A field whose suggestions all fall outside the window is dropped entirely, and a flaw drops out completely if none of its suggestions are in range.

To make this correct regardless of when a request's window is, the per-flaw cache now stores each flaw's raw KPI-relevant data (the suggestion history with per-entry timestamps, plus the current field values) instead of pre-aggregated stats, and re-scores it per request. The date filter is therefore never baked into the cache. New regression tests cover the exact scenario you hit: `test_date_filter_drops_stale_field_but_keeps_recent_one` (one flaw, old `components` suggestion dropped while a recent `impact` suggestion is kept) and `test_date_filters_restrict_by_suggestion_timestamp`.

kdudka (changes requested):
@superbuggy Did you see the content of `kpi_cache.json`?

It stores the whole `aegis_meta` dict and the whole `cvss_scores` array.  Vast majority of the stored fields are never read by Aegis.  This needs to be reworked such that only the needed data is cached.

I would say that the previous data model of the cache was sufficient.  It was only missing the per-field suggestion timestamps.

Moreover, the cache behaves quite unexpectedly I would say:

1. When I query KPI between August 23 and 26, it fetches 715 flaws from OSIDB.
2. When I query KPI between August 27 and 30, it fetches 520 flaws from OSIDB.
3. Then, when I run the first query again, it fetches 715 flaws from OSIDB again!
4. Then, when I run the second query again, it fetches 520 flaws from OSIDB again!

Try this:
```
rm -rf ~/.config/aegis_ai/osidb_bot_kpi

curl -f 'http://localhost:9000/api/v1/analysis/kpi/osidb-bot?changed_after=2026-08-23T00:00:00Z&changed_before=2026-08-26T00:00:00Z'
curl -f 'http://localhost:9000/api/v1/analysis/kpi/osidb-bot?changed_after=2026-08-27T00:00:00Z&changed_before=2026-08-30T00:00:00Z'

curl -f 'http://localhost:9000/api/v1/analysis/kpi/osidb-bot?changed_after=2026-08-23T00:00:00Z&changed_before=2026-08-26T00:00:00Z'
curl -f 'http://localhost:9000/api/v1/analysis/kpi/osidb-bot?changed_after=2026-08-27T00:00:00Z&changed_before=2026-08-30T00:00:00Z'
```

superbuggy:
> @superbuggy Did you see the content of `kpi_cache.json`?
> 
> It stores the whole `aegis_meta` dict and the whole `cvss_scores` array. Vast majority of the stored fields are never read by Aegis. 

Done, the per-suggestion timestamp design behind that change didn't really work out ultimately, nor was there any real gain from having a complicated partial reconciliation caching logic based around that.

src/aegis_ai_web/src/endpoints/bot_kpi.py
```diff
@@ -0,0 +1,803 @@
...
+    so a narrow query (e.g. the last few days) fetches only that window.
+    ``cve_id__isempty=False`` drops flaws without a CVE ID server-side instead of
+    fetching them only to discard them here.
+    """
+    kwargs: dict[str, Any] = {
+        "include_fields": "cve_id,updated_dt",
+        "workflow_state": "DONE",
+        "created_dt_gte": OSIDB_BOT_BIRTHDAY,
+        "cve_id__isempty": False,
+        "limit": 200,
```
kdudka: Please increase `limit` to 1000 like we do in osidb-bot.  When I query CVEs updated since August 1st, it results in ~9500 DONE flaws.  In this case, the change reduces the initial query time from 159s to 52s.

src/aegis_ai_web/src/endpoints/bot_kpi.py
```diff
@@ -0,0 +1,803 @@
...
+    """
+    if field_name == "impact":
+        try:
+            return round(1.0 - score_impact_diff(str(suggested), str(current)), 4)
+        except KeyError:
+            return 1.0
+    if field_name == "_cvss3_vector":
+        score, _reason = score_cvss3_diff(str(suggested), str(current))
+        return round(1.0 - score, 4)
+    return 1.0
```
kdudka, line 90: This might be too coarse for components suggestion.  In cases where a component is converted to lowercase, or a single component is added/removed from the list of components, we might consider returning something between 0.0 and 1.0.  This can also be implemented in the future at the cost of invalidating the cache.

docs/openapi.yml
```diff
@@ -1,5 +1,87 @@
...
+    BotFeatureKPI:
+      description: Per-feature KPI metrics for osidb-bot auto-processing.
+      properties:
+        acceptance_rate:
+          description: Percentage of applied suggestions kept unchanged (0.0-100.0)
+          maximum: 100.0
+          minimum: 0.0
+          title: Acceptance Rate
+          type: number
+        applied:
```
kdudka: `applied` sounds ambiguous to me in this context.  Would `suggested` work better as a name for this field?

docs/openapi.yml
```diff
@@ -1,5 +1,87 @@
...
+            divergence) between modified suggestions and current values, using the
+            same weighted scoring as evals. None when no deviation data is available.
+          title: Avg Suggestion Deviation
+        kept:
+          description: Applied suggestions still unchanged in OSIDB
+          minimum: 0.0
+          title: Kept
+          type: integer
+        modified:
+          description: Applied suggestions that were changed since being applied
```
kdudka: Is it safe to expect that a sum of `kept` and `modified` is equal to `applied`?

This does not always seem to be the case:
```js
% curl --no-progress-meter -f 'http://localhost:9000/api/v1/analysis/kpi/osidb-bot?changed_after=2026-08-01T00:00:00Z&changed_before=2026-09-01T00:00:00Z' | jq '.features["title"]'
{
  "applied": 5381,
  "skipped": 29,
  "kept": 5112,
  "modified": 268,
  "acceptance_rate": 95.0,
  "avg_data_quality": 0.94,
  "avg_confidence": 0.94,
  "avg_suggestion_deviation": 0.05
}
```

Do we know where these discrepancies come from?

src/aegis_ai_web/src/data_models.py
```diff
@@ -331,6 +331,75 @@ def validate_csv_headers(csv_headers: list[str]) -> bool:
...
+    )
+    acceptance_rate: float = Field(
+        ...,
+        ge=0.0,
+        le=100.0,
+        description="Percentage of applied suggestions kept unchanged (0.0-100.0)",
+    )
+    avg_data_quality: float | None = Field(
+        default=None,
+        ge=0.0,
```
kdudka, line 373: Why is `le=1.0` missing here?  This should match the definitions in `AegisFeatureModel`: https://github.com/RedHatProductSecurity/aegis-ai/blob/f3bff3e90f9bb2c821ef4c49c8b91282003c12a9/src/aegis_ai/features/data_models.py#L54-L66

src/aegis_ai_web/src/data_models.py
```diff
@@ -331,6 +331,75 @@ def validate_csv_headers(csv_headers: list[str]) -> bool:
...
+        description=(
+            "Average data_quality score across entries that reported one. "
+            "None when no entry recorded a data_quality value."
+        ),
+    )
+    avg_confidence: float | None = Field(
+        default=None,
+        ge=0.0,
```
kdudka, line 382: Why is `le=1.0` missing here?  This should match the definitions in `AegisFeatureModel`: https://github.com/RedHatProductSecurity/aegis-ai/blob/f3bff3e90f9bb2c821ef4c49c8b91282003c12a9/src/aegis_ai/features/data_models.py#L54-L66

src/aegis_ai_web/src/data_models.py
```diff
@@ -331,6 +331,75 @@ def validate_csv_headers(csv_headers: list[str]) -> bool:
...
+    )
+
+
+class BotKPIResponse(BaseModel):
+    """Response model for osidb-bot KPI metrics endpoint."""
+
+    total_flaws_processed: int = Field(
+        ...,
+        ge=0,
+        description="Total flaws processed by the bot",
```
kdudka: How does the selected time range affect this number?  Does it count flaws processed by osidb-bot where `updated_dt` is within the selected time range?  Or flaws that were processed by osidb-bot within the selected time range?

kdudka (changes requested):
@superbuggy Thanks for the update! I believe this starts to work as expected.  A few comments/suggestions inline...

docs/openapi.yml
superbuggy: Yes that naming is more precise

src/aegis_ai_web/src/endpoints/bot_kpi.py
superbuggy: Done in b5cba0fd. Raised the flaw-index `limit` to 1000 to match osidb-bot.

superbuggy: Done in b5cba0fd. Raised the flaw-index `limit` to 1000 to match osidb-bot.

src/aegis_ai_web/src/endpoints/bot_kpi.py
superbuggy: Done in b8ff38e2. Rather than defer this, `components` now gets a graded 0.0-1.0 deviation via a new `score_components_diff` helper (alongside `score_impact_diff`/`score_cvss3_diff`): names are normalized (lowercased, stripped) and compared with a Jaccard index, so a case-only change scores 0.0 deviation and a single component added/removed from a list of N scores 1/(N+1) instead of a flat 1.0.

docs/openapi.yml
superbuggy: Done in ef95e609. Renamed `applied` to `suggested` across the model, endpoint, and tests, and regenerated the OpenAPI docs.

src/aegis_ai_web/src/data_models.py
superbuggy: Done in ef95e609. Added `le=1.0` to `avg_data_quality` to match `AegisFeatureModel`.

src/aegis_ai_web/src/data_models.py
superbuggy: Done in ef95e609. Added `le=1.0` to `avg_confidence` to match `AegisFeatureModel`.

docs/openapi.yml
superbuggy: No, `kept + modified` is not expected to equal `suggested` (renamed from `applied` in ef95e609). `suggested` counts every AI-Bot suggestion entry, including re-suggestions of the *same* field, whereas `kept`/`modified` reflect one accept-or-modify decision per field, comparing only the *latest* suggestion against the current value. A field re-suggested more than once therefore adds to `suggested` each time but is counted once in `kept`/`modified`, so `suggested >= kept + modified`; the gap is the number of superseded re-suggestions. The updated `suggested` description in the schema now states this relationship explicitly.

src/aegis_ai_web/src/data_models.py
superbuggy: It counts flaws whose suggestions fall *within the selected time range* — filtered by each suggestion's own recorded `timestamp` in `aegis_meta`, not the flaw's `updated_dt`. A bot-processed flaw contributes to `total_flaws_processed` only if at least one of its suggestions is timestamped inside the `changed_after`/`changed_before` window (all bot-processed flaws when no range is given). So a flaw whose `updated_dt` moved last week but whose suggestions were all made months earlier does not count in a "last 7 days" query. I clarified the field description in ef95e609.

src/aegis_ai_web/src/endpoints/bot_kpi.py
kdudka: @superbuggy Thanks for the update!  This is way better than it was.  The problem is that it also makes the comparison case-insensitive:
```py
% PYTHONPATH=./src uv run python
>>> from aegis_ai.features.cve.impact_mappings import score_components_diff
>>> score_components_diff(["kernel", "kernel-rt"], ["kernel"])
0.5
>>> score_components_diff(["Glib"], ["glib"])
1.0
```

Consequently, we get `deviation == 0.0` while `kept == 0` and `modified == 1`, which is inconsistent.

I propose to multiple the score from `score_components_diff()` by `0.9` to reflect this.

docs/openapi.yml
kdudka: @superbuggy When a single field is suggested multiple times, I would count it as a single suggestion.  This way we can always check the integrity of the KPI data (e.g. when the server-side cache gets corrupted).

Note that suggesting a single field multiple times is something that _should not_ happen under normal circumstances.  One would need to run `osidb-bot` with `--force` and even with this option some fields are protected from being overwritten in `suggest.py`:
https://github.com/RedHatProductSecurity/aegis-ai/blob/6a35eebda5c2433262fb49071f4dd26f5d454bd9/src/aegis_ai/osidb_bot/suggest.py#L153
https://github.com/RedHatProductSecurity/aegis-ai/blob/6a35eebda5c2433262fb49071f4dd26f5d454bd9/src/aegis_ai/osidb_bot/suggest.py#L215

kdudka (changes requested):
@superbuggy While trying to rebase this onto `main`, I hit file conflicts on `docs/openapi.{json,yml}`.  I tried to resolve them by running `make generate-openapi` but this ended up with **syntax errors** on two distinct commits from this PR.  This should not happen on a non-draft PR.  I submitted #846 to catch mistakes like this.  Please review.

docs/openapi.yml
superbuggy: I resolved also duplicate skipped counts.

src/aegis_ai_web/src/endpoints/bot_kpi.py
superbuggy: I think the following makes sense: case difference -> `0.1`, variant/suffix difference -> `0.5`, unrelated -> `1.0`... wydt?

superbuggy: a892512a

kdudka: @superbuggy As written above, I would simply multiply the score returned from `score_components_diff()` by `0.9`.  When the function is called, we already know that the initially suggested value was modified later on.  And the proposed one-line fix would ensure that `deviation >= 0.1` will be set in this case.

kdudka: @superbuggy I do not like the solution in a892512afc398402a00624cdc8b1800c3e376633 - it adds complexity and, at the same time, it introduces new problems:
```py
% PYTHONPATH=./src uv run python
>>> from aegis_ai.features.cve.impact_mappings import score_components_diff
>>> score_components_diff(["kernel", "kernel-rt"], ["kernel"])
0.5
>>> score_components_diff(["kernel-rt"], ["kernel"])
0.5
>>> score_components_diff(["kernel-srpm-macros"], ["kernel"])
0.5
>>> score_components_diff(["kernelshark"], ["kernel"])
0.5
```
The last 3 cases should be scored zero.  Please revert.

superbuggy: Reverted in https://github.com/RedHatProductSecurity/aegis-ai/commit/2aa4a933081983ef88020a0f13c911ee30eae626

kdudka (approved):
@superbuggy Looks good, thanks for the update!

PR #746, opened 2026-07-16
refactor(evals): replace raw HTML cache with pre-stripped plaintext

kdudka:
## What

Replace 308 cached raw GitHub commit HTML pages (85 MB) in
`evals/kernel_patch_cache/html/` with pre-stripped plaintext files in
`evals/kernel_patch_cache/text/` (1.9 MB) — a 98% size reduction.

## Why

The only consumer of the cached HTML — `extract_html_features()` — immediately
strips all tags and runs regex feature extraction on the resulting plaintext.
Caching the stripped text instead of raw HTML follows the same approach already
used by `evals/external_references_cache/`, which stores `extracted_text`
rather than raw HTML pages.

Savings:

- Working-tree size: 85 MB → 1.9 MB (98% reduction)
- Git objects removed: 308 files, 449 276 deleted lines
- Feature extraction: verified identical across all 308 files
  (`strip_html()` is idempotent on already-stripped text)

## How to Test

- `make check` — format, lint, type-check all pass
- `make test` — 227 unit tests pass
- `make test-web` — 152 web tests pass
- Feature parity verified: `extract_html_features()` produces identical
  results for all 308 commits whether fed raw HTML or pre-stripped text

## Related Tickets

Related: https://redhat.atlassian.net/browse/AEGIS-461

🤖 Generated with [Claude Code](https://claude.com/claude-code)


evals/utils/populate_kernel_cve_cache.py
```diff
@@ -229,17 +230,27 @@ async def _populate_cve_context(cve_ids: list[str]) -> None:
 
 
 async def _fetch_artifacts(hashes: set[str], kind: str) -> None:
-    """Fetch artifacts (patches or HTML) and write to the cache directory."""
+    """Fetch artifacts (patches or HTML) and write to the cache directory.
+
+    For HTML artifacts, also writes a stripped plaintext version to
+    ``text/{hash}.txt`` (committed to git instead of raw HTML).
+    """
```
superbuggy, line 237: Is the intent here to eventually move away from the HTML cache, no opinion just curious. I suppose we could specify our own XML format to impose structure, but I like the idea of moving away from HTML.

superbuggy (approved):

evals/utils/populate_kernel_cve_cache.py
kdudka: This PR only optimizes the caching so that we do not maintain 85 MB of data in git out of which 83 MB is not visible to evals anyway.  If we wanted to use a structured format, it would likely be JSON, which is already used by other evals fixtures.  But let's implement the git access first to see whether the HTML processing can be dropped altogether.

PR #804, opened 2026-08-11
feat(kernel): centralize vulns.git access, unblock git.kernel.org network dependency (AEGIS-484)

superbuggy:
## What

Centralizes the duplicated git clone/pull logic for the upstream `linux/security/vulns` repo (previously three independent, drifting implementations) into a single `sync_vulns_repo()` in `src/aegis_ai/kernel_vulns_repo.py`, and switches the default source from `git.kernel.org` to `kernel.googlesource.com`, which mirrors the same repo at the same path under a different host. Also reorders the kernel classifier's per-commit patch-fetch template list (`PATCH_URL_TEMPLATES`) to try `gregkh/linux` (GitHub) before `git.kernel.org`, matching the ordering `HTML_COMMIT_URL_TEMPLATES` already used, and extracts a shared `_fetch_first_match()` helper to remove the duplicated fetch loops.

## Why

`git.kernel.org` is blocked by Akamai gateways on some Red Hat production networks (MOSAIC), and the long-term fix is to stop depending on it where a drop-in alternative exists. `kernel.googlesource.com` mirrors `linux/security/vulns` at the identical path, so the vulns.git clone (used by the kernel CVE lookup tool, the classifier's training pipeline, and a standalone prediction CLI) can move over directly. The classifier's raw-patch/commit-HTML fetch hits `git.kernel.org`'s cgit web UI, which `kernel.googlesource.com` (Gitiles) doesn't serve the same way, so that path is fixed by reordering to the already-working GitHub mirror instead of a host swap.

## How to Test

- `make check` (ruff format/lint + ty type check)
- `make test` — full unit suite, including new coverage in `tests/test_kernel_vulns_repo.py` (clone/pull/failure paths, `describe_git_error` fallback) and `tests/test_kernel_classifier_fetch.py` (`_fetch_first_match` success/fallthrough/all-fail/exception-continue paths)
- Manual: delete any local `config_dir/kernel_cves/linux_security_vulns` clone and exercise the kernel CVE lookup tool; confirm it clones from `kernel.googlesource.com` and kernel CVE classification still proceeds end-to-end

## Related Tickets
Related: https://redhat.atlassian.net/browse/AEGIS-484

## Summary by Sourcery

Centralize kernel vulnerability repository access and improve kernel commit fetching reliability across production network environments.

New Features:
- Centralize Linux kernel vulnerabilities repository synchronization behind a shared clone-or-pull utility using kernel.googlesource.com by default.

Bug Fixes:
- Avoid dependence on blocked git.kernel.org access for vulnerabilities metadata and improve recovery from incomplete clones while retaining usable stale data after pull failures.
- Reject invalid kernel commit hashes before repository operations to prevent malformed revisions from being processed.

Enhancements:
- Prioritize the GitHub gregkh/linux mirror for kernel patch retrieval and share fallback-fetch behavior across patch and commit HTML fetching.

Tests:
- Add coverage for vulnerabilities repository cloning, pulling, failures, incomplete clone recovery, error reporting, and fetch fallback behavior.


superbuggy:
Addressed the blocking Sourcery security check (check_run_id 94136969993): `subprocess.run` in `scrape_from_commits()` (cve_predictor.py:263-269) used `commit_hash` without a static string, flagged as a possible command-injection audit target.

Since the command is passed as an argv list (never `shell=True`), classic shell injection isn't possible here, but an unvalidated value starting with `-` could still be misread by `git` as an option rather than a revision (argument injection). `main()` already validated commit hashes as 40-char hex before constructing `CVEPredictor`, but that guard only lived at the CLI layer — the class itself didn't enforce it. Extracted the regex to a shared `COMMIT_HASH_RE` constant and added validation at the point of use in `scrape_from_commits()`, which also covers `_ensure_commit_exists()`'s `git cat-file`/`git fetch` calls since it runs first in the same per-commit loop.

Fixed in c8947ccd.

superbuggy:
@sourcery-ai the same finding is still flagged on the latest commit (now at cve_predictor.py:275-281, shifted by the fix in c8947ccd) — same `git show` call, same `commit_hash` variable.

To close the loop: `commit_hash` is validated against `COMMIT_HASH_RE = re.compile(r"^[0-9a-fA-F]{40}$")` immediately upon entering the per-commit loop (cve_predictor.py:263-265), before any of the `git` subprocess calls in this method or in `_ensure_commit_exists()` run. Since the command is passed as an argv list (no `shell=True`), shell metacharacter injection isn't possible regardless; the actual risk this check is auditing for — a value that could be misread by `git` as an option rather than a revision — is closed by the hex-only regex, since `-` is not a valid character in the allowed set. `shlex.quote()`/`shlex.escape()` wouldn't add anything here (it protects shell string construction, not argv-list option/argument confusion) and isn't the right tool for a non-shell subprocess call.

@sourcery-ai resolve

superbuggy:
Addressed the remaining actionable review findings (commits e7fcf747, 30b57f66):

**Fixed**
- **CodeRabbit** — `cve_predictor.py` commit-hash validation now uses `COMMIT_HASH_RE.fullmatch()` instead of `.match()`. With `.match()`, the anchored `^…$` still accepts a 40-hex value with a trailing newline (`$` tolerates it) before it flows into a `git` argv list; `fullmatch` rejects it. Dropped the now-redundant anchors.
- **CodeRabbit (Major)** — routed an existing-but-incomplete clone (directory present, no `.git`) through the rmtree+re-clone recovery path in `check_repository_setup()`. Previously that only ran when `missing_repos` was non-empty, so an incomplete repo left alongside a valid one was logged but never recovered, and later broke the symlink/scrape step.
- **Sourcery** — hoisted the hardcoded `_fetch_first_match` `min_length` values (100 / 200) into `_PATCH_MIN_LENGTH` / `_HTML_MIN_LENGTH`, alongside the existing `_*_SIZE_LIMIT` constants.

**Intentionally skipped** (already handled or opinionated)
- Sourcery "log the `False`/stale case in callers" — `kernel_cves.setup` and `ensure_repo` already act on the return, and `sync_vulns_repo` logs the stale-data warning internally.
- Sourcery "`ensure_repo` always says *Failed to clone*" — accurate as written: `sync_vulns_repo` only *raises* on clone failure; pull failures return `False`.
- Sourcery test-assertion nit — already applied to `test_sync_vulns_repo_reclones_when_existing_path_is_not_a_git_repo`.
- CodeRabbit atomic clone (temp-dir + rename) — `sync_vulns_repo` already removes and re-clones incomplete clones; the atomic variant is an enhancement, not a correctness fix.
- CodeRabbit "always pull a valid existing vulns repo in `cve_predictor`" — that's a behavior change for a standalone offline prediction CLI (the large `linux_kernel_repo` is deliberately never re-pulled either), so left as-is.

`make all` passes (ruff + ty, unit tests, web tests, OpenAPI check).

tests/test_kernel_vulns_repo.py
```diff
@@ -0,0 +1,125 @@
...
+
+    def fake_run(cmd, **kwargs):
+        calls.append(cmd)
+        return MagicMock(returncode=0)
+
+    monkeypatch.setattr(subprocess, "run", fake_run)
+
+    result = sync_vulns_repo(repo_path, "https://example.com/vulns")
+
+    assert result is True
```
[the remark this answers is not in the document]
superbuggy: Already addressed on the branch (commit 33e1a811). The misleading `assert not repo_path.exists()` is gone; the re-clone regression test (`test_sync_vulns_repo_reclones_when_existing_path_is_not_a_git_repo`) now stubs `git clone` to create the repo dir + `.git` and asserts real postconditions (result True, `repo_path.exists()`, `(repo_path / '.git').exists()`, old contents removed, and the expected clone argv) — matching this suggestion. Resolving.

src/aegis_ai_ml/src/classifier/kernel-cve-impact-classifier/cve_predictor.py
```diff
@@ -10,6 +10,7 @@
 import json
 import logging
 import re
+import shlex
```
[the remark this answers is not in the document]
superbuggy: Already addressed on the branch (commits 602b7fbc / e7fcf747). Commit-hash validation now uses `COMMIT_HASH_RE.fullmatch(...)` in both `scrape_from_commits()` and `main()`, so a trailing newline is rejected. No `.match()` remains for hash validation. Resolving.

src/aegis_ai/kernel_vulns_repo.py
```diff
@@ -0,0 +1,97 @@
...
+    if the initial clone fails — callers decide how to react to a repo that
+    was never successfully established. A failed pull on an already-working
+    repo is logged and swallowed so callers keep working with stale data
+    rather than losing an otherwise-usable clone.
+
+    Returns True if the repo is confirmed current (clone or pull succeeded),
+    False if a pull failed and the existing clone was left as-is.
+    """
+    if repo_path.exists() and not (repo_path / ".git").exists():
+        logger.warning("Removing incomplete clone at %s and re-cloning", repo_path)
```
kdudka: The logic seems wrong.  Why do you think an incomplete `git clone` is signaled by missing `.git`?

src/aegis_ai/kernel_vulns_repo.py
```diff
@@ -0,0 +1,97 @@
...
+        shutil.rmtree(repo_path)
+
+    if not repo_path.exists():
```
kdudka, line 69: This does not seem safe wrt. parallel execution.  The TOCTOU between `.exists()` and `.mkdir()` is less likely to hit in practice. But even running two `git pull` processes (or `git clone` vs. `git pull`) in parallel on a single git repo is not supported IMO.  This could lead to difficult to debug failures if we moved `$XDG_CONFIG_HOME` to a shared directory.  I think we should at least warn about this limitation.

src/aegis_ai/kernel_vulns_repo.py
```diff
@@ -0,0 +1,97 @@
...
+    Raises ``subprocess.CalledProcessError`` or ``subprocess.TimeoutExpired``
+    if the initial clone fails — callers decide how to react to a repo that
+    was never successfully established. A failed pull on an already-working
+    repo is logged and swallowed so callers keep working with stale data
+    rather than losing an otherwise-usable clone.
+
+    Returns True if the repo is confirmed current (clone or pull succeeded),
+    False if a pull failed and the existing clone was left as-is.
+    """
+    if repo_path.exists() and not (repo_path / ".git").exists():
```
kdudka, line 65: The fact that `.git` exists does not imply a completed `git clone` on its own.  It would be better to check for the presence of `.git/index` instead.

kdudka (changes requested):

src/aegis_ai/kernel_vulns_repo.py
```diff
@@ -0,0 +1,113 @@
...
+        logger.warning("Removing incomplete clone at %s and re-cloning", repo_path)
+        shutil.rmtree(repo_path)
+
+    # Concurrency limitation: this function is not safe to run against the same
+    # repo_path from two processes at once. The exists()/mkdir()/clone sequence
+    # below is a TOCTOU race, and running two git pull/clone processes against
+    # the same working tree is unsupported by git and can corrupt the clone —
+    # especially if XDG_CONFIG_HOME points at a shared directory. Callers must
+    # serialize access to a given repo_path themselves.
+    logger.warning(
```
kdudka, line 79: I meant a warning in `docstring` or documentation, not a run-time warning on each git access.  That would be excessively disruptive on production.

kdudka (commented):
This needs to rebased onto `main` because of commits 89c6429df2aad757bd1a172e871165b1d74cf14f and cba2b08e1c747e0d817715323a707c5f0fdd647d, which have been merged meanwhile.  Sorry for the extra work!

kdudka (approved):
@superbuggy Looks good.  Thanks for the update!

PR #835, opened 2026-09-02
Add KPI filters and detail enrichment for component feedback

superbuggy:
## What

> In place of #700

  Extend `GET /api/v1/analysis/kpi/cve` with component feedback filtering and optional
  detail enrichment (`src/aegis_ai_web/src/endpoints/kpi.py`, `main.py`,
  `data_models.py`). Adds `cve_id`, `source_component`, `multiple_source_components`,
  and `detail` query params; refactors the KPI pipeline to normalize manual and
  programmatic feedback, apply filters, and compute component diffs
  (accepted/rejected/added). `detail=true` returns enriched `KPIEntry` fields; responses
  use `response_model_exclude_none=True` for compact payloads.

  Incorporates review feedback: `email` is dropped from the response model (PII/PIA);
  `source_component` aliasing is one-way so `suggest-affected-components` keeps its
  original scope; component matching is case-sensitive and preserves source ordering;
  and component filters apply only to component features so `feature=all` excludes
  non-component features.

  ## Why

  Analysts need to query KPI data by CVE, suggested component, and multi-component
  suggestions, and inspect component diffs without post-processing raw feedback logs —
  while keeping submitter PII out of the API contract.

  ## How to Test

  ```bash
  uv run pytest src/aegis_ai_web/tests/test_kpi.py -q
  make check
```

> After staging deploy:
```bash
  curl --negotiate -u : -f \
    'https://aegis.prodsec.redhat.com/api/v1/analysis/kpi/cve?feature=source_component&s
  ource_component=kernel&detail=true'
```
## Related Tickets

Related: https://redhat.atlassian.net/browse/AEGIS-405

## Summary by Sourcery

Enhance CVE KPI analysis with component-focused filtering and consistent feedback reporting across data sources.

New Features:
- Add CVE, suggested-component, and multi-component filters to the CVE KPI endpoint.
- Unify manual and programmatic feedback under canonical public feature names.
- Support component-aware KPI filtering with case-sensitive matching and multi-component selection.

Bug Fixes:
- Prevent submitter email and other raw feature aliases from leaking into KPI responses.
- Exclude non-component features when component filters are applied across all features.

Enhancements:
- Normalize feedback data from manual and programmatic sources before calculating KPI metrics.
- Preserve component ordering while cleaning and de-duplicating suggested component lists.
- Expose the updated KPI query parameters and feature names in the API documentation.

Documentation:
- Document the new KPI filters and canonical feature naming in the changelog and OpenAPI specifications.

Tests:
- Add unit and API coverage for KPI filtering, feature alias normalization, mixed feedback sources, response shape, and invalid CVE identifiers.


src/aegis_ai_web/src/endpoints/kpi.py
```diff
@@ -84,66 +115,233 @@ def _deduplicate_programmatic_feedback(
...
+    accepted_components, rejected, added = component_diff(suggested, submitted)
+
+    return KPIEntry(
+        datetime=entry["datetime"],
+        accepted=entry["accepted"],
+        aegis_version=entry["aegis_version"],
+        cve_id=entry.get("cve_id") or None,
+        feedback_source=entry.get("feedback_source"),
+        suggested_components=suggested or None,
+        submitted_components=submitted or None,
```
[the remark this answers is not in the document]
superbuggy: Fixed in https://github.com/RedHatProductSecurity/aegis-ai/pull/835/commits/2b05809b50ab3e0d19ecd96c37f937d43d189b00

docs/CHANGELOG.md
```diff
@@ -4,6 +4,14 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Added
+- KPI endpoint filters for component feedback: `cve_id`, `source_component`, and `multiple_source_components` query parameters on `GET /api/v1/analysis/kpi/cve`
+- `detail=true` on the KPI endpoint returns CVE and component diff fields (`suggested_components`, `accepted_components`, `rejected_suggestions`, `added_components`)
+- `source_component` feature alias on the KPI endpoint maps to both `source_component` and `suggest-affected-components` feedback logs (one-way; querying `suggest-affected-components` keeps its original scope). Component matching is case-sensitive.
```
kdudka: The aliasing does not seem to work as expected.  When I query `feature=all` on the prod data, I get:

- both `cve_description` and `suggest-description`
- both `source_component` and `suggest-affected-components`

I would expect those to be unified in the response:
```js
% curl --no-progress-meter -f 'http://localhost:9000/api/v1/analysis/kpi/cve?feature=all' | jq keys 
[
  "cve_description",
  "source_component",
  "suggest-affected-components",
  "suggest-cvss",
  "suggest-cwe",
  "suggest-description",
  "suggest-impact",
  "suggest-mitigation",
  "suggest-statement",
  "suggest-title"
]
```

src/aegis_ai_web/src/data_models.py
```diff
@@ -347,6 +349,34 @@ class KPIEntry(BaseModel):
...
+        description="Suggested components also present in submission (included when detail=true)",
+    )
+    rejected_suggestions: list[str] | None = Field(
+        default=None,
+        description="Suggested components not present in submission (included when detail=true)",
+    )
+    added_components: list[str] | None = Field(
+        default=None,
+        description="Submitted components not in the suggestion (included when detail=true)",
+    )
```
kdudka: The model seems a bit inconsistent.  While most of the fields are generic, `suggested_components`, `submitted_components`, `accepted_components`, and `added_components` are specific to components suggestion and they do not make much sense for other KPI entries.  This should be reworked to keep the model sufficiently simple, generic, and easy to reuse/extend.

kdudka (changes requested):

src/aegis_ai_web/src/data_models.py
superbuggy: > `sted_components`, `submitted_components`, `accepted_components`, and `added_components` are specific to components suggestion and they do not make much sense for other KPI entries.

I think the original author's intent was to capture some of these more granular differences and I preserved that approach initially, but I am not sure it is valuable to add these fields. The data that would go in these proposed fields could be queried when necessary, should the need actually arise. I think I will drop these fields, if they are able to be reconstructed with an OSIDB query.

src/aegis_ai_web/src/endpoints/kpi.py
```diff
@@ -84,43 +99,177 @@ def _deduplicate_programmatic_feedback(
...
+        ):
+            entries.append(normalized)
+
+    deduped_programmatic = _deduplicate_programmatic_feedback(
+        programmatic_feedback_logger.read()
+    )
+    for raw in deduped_programmatic:
+        if not raw.get("feature"):
+            continue
+        normalized = normalize_programmatic_entry(raw)
```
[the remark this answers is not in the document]
superbuggy: Fixed in 2f09154f — `_deduplicate_programmatic_feedback()` now builds its key with `canonical_feature()`, so `source_component` and `suggest-affected-components` rows for the same CVE collapse to one sample. Added `test_programmatic_aliases_dedupe_by_canonical_feature` covering both aliases for one CVE and asserting a single sample at 100% acceptance.

src/aegis_ai_web/src/main.py
```diff
@@ -751,13 +774,19 @@ async def cve_kpi(
 
     **Example:**
     ```
-    GET /api/v1/analysis/kpi/cve?feature=suggest-impact&order=desc
+    GET /api/v1/analysis/kpi/cve?feature=source_component&order=desc
+    GET /api/v1/analysis/kpi/cve?feature=source_component&source_component=kernel
```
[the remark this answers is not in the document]
superbuggy: Fixed in 2f09154f — `get_cve_kpi` now keys the single-feature response with `canonical_feature(feature)`, so querying the raw alias `source_component` responds under `suggest-affected-components` (matching the `feature=all` behavior). Raw aliases are still accepted for request filtering. Tests updated to assert the canonical response key.

src/aegis_ai_web/src/data_models.py
superbuggy: Ended up dropping the fields to simplify

superbuggy: Ended up dropping the fields to simplify

src/aegis_ai_web/src/main.py
```diff
@@ -726,14 +726,34 @@ async def component_analysis(
 async def cve_kpi(
     feature: str = Query(
         ...,
-        description="Feature name to filter entries by. Valid values include: 'suggest-impact', 'suggest-cwe', 'suggest-description', 'suggest-statement', 'identify-pii', 'cvss-diff-explainer', or 'all' to get KPIs for all features.",
-        examples=["suggest-impact", "suggest-cwe", "suggest-description", "all"],
+        description="Feature name to filter entries by. Valid values include: 'suggest-impact', 'suggest-cwe', 'suggest-description', 'suggest-statement', 'identify-pii', 'cvss-diff-explainer', 'source_component', 'suggest-affected-components', or 'all' to get KPIs for all features.",
```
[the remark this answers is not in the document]
kdudka: The `query-affected-components` feature is experimental: https://github.com/RedHatProductSecurity/aegis-ai/blob/ebea923072df59ca1102895f29d59d67916e91db/src/aegis_ai/features/cve/data_models.py#L73

kdudka (changes requested):
@superbuggy The implementation of the `detail=true` flag seems to have been lost in the last rebase.  Now I am not able to get even `cve_id` field in the returned entries.

superbuggy:
> The implementation of the `detail=true` flag seems to have been lost in the last rebase. Now I am not able to get even `cve_id` field in the returned entries.

Fixed now.

src/aegis_ai_web/src/main.py
```diff
@@ -730,14 +731,39 @@ async def component_analysis(
 async def cve_kpi(
     feature: str = Query(
         ...,
-        description="Feature name to filter entries by. Valid values include: 'suggest-impact', 'suggest-cwe', 'suggest-description', 'suggest-statement', 'identify-pii', 'cvss-diff-explainer', 'quality-review', or 'all' to get KPIs for all features.",
-        examples=["suggest-impact", "suggest-cwe", "suggest-description", "all"],
+        description="Feature name to filter entries by. Valid values include: 'suggest-impact', 'suggest-cwe', 'suggest-description', 'suggest-statement', 'identify-pii', 'cvss-diff-explainer', 'quality-review', 'source_component', 'suggest-affected-components', or 'all' to get KPIs for all features.",
+        examples=[
+            "suggest-impact",
+            "quality-review",
+            "source_component",
```
kdudka, line 738: `source_component` should not be in `description` and `examples` because it is not a name of an Aegis feature.  We do not want to advertise it in the supported API as it only makes it look inconsistent and confusing.

src/aegis_ai_web/src/data_models.py
```diff
@@ -421,11 +421,43 @@ class BotKPIResponse(BaseModel):
...
+        description="Suggested components also present in submission",
+    )
+    rejected_suggestions: list[str] | None = Field(
+        default=None,
+        description="Suggested components not present in submission",
+    )
+    added_components: list[str] | None = Field(
+        default=None,
+        description="Submitted components not in the suggestion",
+    )
```
kdudka, line 451: I still do not understand what each of the fields means in practice.  I did a quick check with a random component:
```sh
% curl --no-progress-meter 'http://localhost:9000/api/v1/analysis/kpi/cve?feature=all&source_component=curl&detail=true' | jq
```
```json
{
  "suggest-affected-components": {
    "acceptance_percentage": 0.0,
    "entries": [
      {
        "datetime": "2026-05-29 08:10:12.739",
        "accepted": false,
        "aegis_version": "0.7.2",
        "cve_id": "CVE-2026-49129",
        "feedback_source": "manual",
        "components": {
          "suggested_components": [
            "mpd",
            "curl"
          ],
          "rejected_suggestions": [
            "mpd",
            "curl"
          ]
        }
      }
    ]
  }
}
```

`suggested_components` sounds clear.  But the content of `rejected_suggestions` is misleading because it seems that both the suggested components were wrong.  However, in fact, only the `curl` component was wrong whereas `mpd` was the correct component.  The corresponding entry in `feedback.csv` was:
```csv
2026-05-29 08:10:12.739,source_component,CVE-2026-49129,XXX,"[""mpd"",""curl""]",,0ms,false,"Vulnerability is in MPD ()MusicPlayerDaemon), which we do not ship. It wrongly also added ""curl"" as component which has nothing to do with this MPD vulnerability",0.7.2
```

Long story short, I find the interface over-complicated and not practically useful.  I would only keep `suggested_components` and `submitted_components`.  The latter would only be available for entries from programmatic feedback.

src/aegis_ai_web/src/main.py
```diff
@@ -756,13 +786,21 @@ async def cve_kpi(
 
     **Example:**
     ```
-    GET /api/v1/analysis/kpi/cve?feature=suggest-impact&order=desc
+    GET /api/v1/analysis/kpi/cve?feature=source_component&order=desc
+    GET /api/v1/analysis/kpi/cve?feature=source_component&source_component=kernel
+    GET /api/v1/analysis/kpi/cve?feature=source_component&cve_id=CVE-2025-1234
+    GET /api/v1/analysis/kpi/cve?feature=source_component&detail=true&source_component=kernel
```
kdudka, line 792: Please use `feature=suggest-affected-components` in the examples to make it consistent with the output data:
```
% curl --no-progress-meter -f 'http://localhost:9000/api/v1/analysis/kpi/cve?feature=all' | jq keys
[
  "suggest-affected-components",
  "suggest-cvss",
  "suggest-cwe",
  "suggest-description",
  "suggest-impact",
  "suggest-mitigation",
  "suggest-statement",
  "suggest-title"
```

kdudka (changes requested):

src/aegis_ai_web/src/data_models.py
superbuggy: > Long story short, I find the interface over-complicated and not practically useful. I would only keep suggested_components and submitted_components. The latter would only be available for entries from programmatic feedback.

Agreed, it is now simplified from the original author's implementation.

docs/openapi.json
```diff
@@ -385,8 +385,45 @@
         "title": "HTTPValidationError",
         "type": "object"
       },
+      "KPIComponentDetails": {
+        "description": "Component-level feedback diff for a KPI entry.\n\nPopulated only for component features when ``detail=true`` is passed to the\nKPI endpoint.",
+        "properties": {
+          "submitted_components": {
```
kdudka: @superbuggy I was unable to test this because there was no such data in `programmatic_feedback.csv`, I am afraid because OSIM does not send programmatic feedback when the suggested components are overridden by the user.  Were you able to test this on some real-world data?

superbuggy: Ah no that's right, we had a parallel task for OSIM that this must have been based on, I've reworked this to use aegis_meta instead but need to test and consider issues with the approach

superbuggy: Dropping the `submitted_components` field for now.

kdudka (approved):
@superbuggy Looks good.  Thanks for the update!

PR #849, opened 2026-09-08
feat: build system toolset for binary RPM enumeration

kdudka:
## What

Add a pydantic-ai toolset that lets the LLM agent look up binary RPM
subpackage names produced by a source RPM in a product stream, using
Deptopia + Brew/Koji APIs. Binary RPM data is pre-computed before the
LLM prompt to avoid per-affect tool call timeouts, then injected into
the affects context so the LLM can identify binary-level affectedness.

## Why

`SuggestAffectedPackages` works at the source RPM level but cannot tell
which specific binaries within a source package are affected. The build
system toolset bridges this gap by resolving source-to-binary RPM
mappings, enabling more precise `affected_subcomponents` suggestions.

## How to Test

```bash
# run all checks
make all

# run evals with build system tool enabled (uses committed cache)
AEGIS_USE_BUILD_SYSTEM_TOOL_CONTEXT=true make eval

# manual test (requires VPN for live Brew/Deptopia)
AEGIS_USE_BUILD_SYSTEM_TOOL_CONTEXT=true uv run aegis suggest-affected-packages CVE-2025-49175
```

## Related Tickets

Related: https://redhat.atlassian.net/browse/AEGIS-491


kdudka:
I resubmitted this as #850 because the target branch had been set incorrectly.

PR #856, opened 2026-09-10
feat: add `component` filter to `kpi/osidb-bot` endpoint

superbuggy:
## Summary
Adds an optional `component` query parameter to the osidb-bot KPI endpoint (`GET /api/v1/analysis/kpi/osidb-bot`), letting callers scope KPI metrics to flaws affecting a specific component — e.g. `?component=kernel`. It composes with the existing `changed_after`/`changed_before` date filters.

## Changes
- **API** (`main.py`): new `component` query param, documented in the endpoint's OpenAPI schema. An empty `?component=` is normalized to "no filter" (`component or None`), matching how the date params are treated.
- **Query** (`bot_kpi.py`): `component` is threaded through `get_osidb_bot_kpi` → `_fetch_with_cache` → `_fetch_flaw_index` and pushed to OSIDB as the flaw-level `components` filter (the field the KPI scores against, so selection membership matches what is measured).
- **Cache correctness**: the shared `kpi_cache.json` holds every component's flaws, so a component request:
  - forces `index_is_complete=False`, so `_reconcile_flaws` never prunes the shared cache down to the component subset;
  - scopes aggregation via `to_kpi_result(only_cve_ids=…)` so only the component's flaws are counted.
- **Date semantics**: date bounds are deliberately left *off* the component-membership index. Bounds filter on a flaw's `updated_dt`, but the KPI scopes dates per *suggestion timestamp* — so a flaw edited after the window whose suggestion is in-window stays in the membership set and is date-filtered during scoring. This keeps `?component=X&changed_before=Y` consistent with the date-only path rather than undercounting.

## Testing
- `_fetch_flaw_index` pushes the `components` filter when a component is given, and omits it otherwise.
- A component request scopes the result to that component and retains other components' flaws in the shared cache (no prune, byte-identical cache file).
- The endpoint pushes the component down to the OSIDB query.
- Regression: `component` + `changed_before` counts an in-window suggestion on a later-edited flaw (per-suggestion date scoping, not `updated_dt`).
- Full suite green: `make check` (ruff format + lint + ty), `make test` (269 passed), `make test-web` (248 passed).

## Notes
- `docs/CHANGELOG.md` not updated — it has no "Unreleased" section and entries are version-bound; happy to add one under the appropriate release.

## Related Tickets
Related: https://redhat.atlassian.net/browse/AEGIS-494

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by Sourcery

Add component-scoped KPI reporting while maintaining complete shared-cache reconciliation and consistent date-filter behavior.

New Features:
- Add an optional component filter to scope osidb-bot KPI metrics to flaws affecting a specified component.

Bug Fixes:
- Preserve correct per-suggestion date filtering for component-scoped KPI requests, including flaws updated outside the requested date range.

Enhancements:
- Separate cache reconciliation from request-specific selection so component and date filters do not prune or invalidate shared cached data.
- Improve concurrent cache reconciliation to preserve newer updates and entries created after a request's snapshot.

Documentation:
- Document the new component query parameter in the endpoint's OpenAPI specification.

Tests:
- Add coverage for component query propagation, result scoping, cache retention, date semantics, and concurrent cache reconciliation.

[AEGIS-494]: https://redhat.atlassian.net/browse/AEGIS-494?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/aegis_ai_web/src/main.py
```diff
@@ -820,6 +820,10 @@ async def osidb_bot_kpi(
         default=None,
         description="Only include flaws updated before this ISO 8601 datetime.",
     ),
+    component: str | None = Query(
```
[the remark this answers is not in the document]
superbuggy: Verified this suggestion does not apply to this repo — the opposite is true, so I'm not adding the suppression.

This project uses ruff'\''s **default** rule set (no `select` in `[tool.ruff.lint]`; only `ignore = [BLE001, LOG015]`), so **B008 (flake8-bugbear) is not enabled** and cannot flag this line.

Running the lint gate (`make check` → `uv run ruff check`) on the current code passes clean:

```
uv run ruff check
All checks passed!
```

Adding `# noqa: B008` here actually **breaks** the gate, because the directive is unused and ruff reports RUF100:

```
RUF100 [*] Unused `noqa` directive (unused: `B008`)
   --> src/aegis_ai_web/src/main.py:823:37
Found 1 error.
```

So the `component` parameter is correct as-is with no suppression. Resolving.

src/aegis_ai_web/src/endpoints/bot_kpi.py
```diff
@@ -421,6 +424,8 @@ def _fetch_flaw_index(
         kwargs["updated_dt_gte"] = changed_after
     if changed_before is not None:
         kwargs["updated_dt_lte"] = changed_before
+    if component is not None:
+        kwargs["affects__ps_component"] = component
```
kdudka, line 428: I believe that the correct search predicate is `kwargs["components"] = component`.

src/aegis_ai_web/src/endpoints/bot_kpi.py
```diff
@@ -763,9 +794,12 @@ def _fetch_with_cache(
...
     # bot birthday); only then can a cached flaw's absence be read as a departure
-    # and pruned. A date-bounded index is just a window, so absence there is
-    # ambiguous and nothing is pruned (see _reconcile_flaws).
-    index_is_complete = changed_after is None and changed_before is None
+    # and pruned. A date-bounded or component-scoped index is only a subset of the
+    # shared cache, so absence there is ambiguous and nothing is pruned (see
+    # _reconcile_flaws).
+    index_is_complete = (
+        changed_after is None and changed_before is None and component is None
+    )
```
kdudka, line 802: I can see that cache invalidation is completely wrong (and this change does not make it any better).  We should not invalidate any cache entries based on query parameters.  Query parameters should only drive selection of CVEs to look for in the cache.  A cache entry (not the entire cache etc.) should only be invalidated when its `updated_dt` field is behind  OSIDB.

kdudka (changes requested):

src/aegis_ai_web/src/endpoints/bot_kpi.py
```diff
@@ -776,22 +771,30 @@ def _fetch_with_cache(
         with _cache_handler() as handler:
             latest = handler.read()
             existing = latest.flaws if latest is not None else cached_flaws
-            entry = BotKPICacheEntry(
-                flaws=_reconcile_flaws(
-                    existing, fetched, index, index_is_complete=index_is_complete
-                )
-            )
+            entry = BotKPICacheEntry(flaws=_reconcile_flaws(existing, fetched, index))
```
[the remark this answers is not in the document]
superbuggy: Fixed in cd2c6ee8. Confirmed the race: the index and batch fetch are captured before the cache lock, but `_reconcile_flaws` pruned/merged against `existing` (re-read under the lock), so a flaw a concurrent request inserted between our index fetch and lock acquisition got pruned (absent from our stale index), and a fetch on an old watermark could clobber a newer concurrent write. The recent refactor to a single always-complete index made this apply to every request, not just unfiltered ones.

Fix: reconcile relative to the pre-lock snapshot (`cached_flaws`), not `existing` — prune only flaws that were in our snapshot but left the index, and apply a fetched entry only when the live watermark still matches the snapshot. Kept fetches outside the lock (moving them in would serialize all requests behind slow OSIDB queries, which the design deliberately avoids). Both divergences self-heal on the next request. Added `TestReconcileFlaws` covering all three cases.

src/aegis_ai_web/src/endpoints/bot_kpi.py
```diff
@@ -417,10 +423,8 @@ def _fetch_flaw_index(
         "cve_id__isempty": False,
         "limit": 1000,
     }
-    if changed_after is not None:
-        kwargs["updated_dt_gte"] = changed_after
-    if changed_before is not None:
-        kwargs["updated_dt_lte"] = changed_before
+    if component is not None:
+        kwargs["components"] = component
```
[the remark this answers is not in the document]
superbuggy: Not changing this — using `components` is deliberate. The KPI scores the *flaw-level* `components` field (`BOT_KPI_FIELDS`, `_compute_deviation` at bot_kpi.py:84), so the selection filter must match what is scored. Filtering on `affects__ps_component` would select flaws whose flaw-level `components` never mentions the component, and scoring them would not attribute anything to it — inconsistent membership vs. measurement. This is asserted on purpose in `tests/test_bot_kpi_cache.py:247-249` ("not the affects-level ps_component"). The PR description still says `affects__ps_component`; that text is stale from the initial commit — the later refactor switched to `components` intentionally. I will update the PR description.

src/aegis_ai_web/src/endpoints/bot_kpi.py
superbuggy: Reworked the cache invalidation to decoupled from query.

src/aegis_ai_web/src/endpoints/bot_kpi.py
```diff
@@ -687,31 +699,42 @@ def _reconcile_flaws(
...
     """
-    if index_is_complete:
-        merged = {cve_id: flaw for cve_id, flaw in existing.items() if cve_id in index}
-    else:
-        merged = dict(existing)
-    merged.update(fetched)
+    departed = set(snapshot) - set(index)
+    merged = {
+        cve_id: flaw for cve_id, flaw in existing.items() if cve_id not in departed
+    }
```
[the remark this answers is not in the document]
superbuggy: Removed this reconciliation logic, no longer relevant

src/aegis_ai_web/src/endpoints/bot_kpi.py
```diff
@@ -267,13 +270,38 @@ def _score_records(
     return stats
 
 
+def _timestamp_in_range(
```
kdudka, line 273: Sorry but this seems still wrong and over complicated.  Why do we need `_timestamp_in_range()`?
The initial OSIDB search, which should be based on the query params, already gives us a list of CVEs within the specified time range.  There is no need to check the range locally at all.

The only thing we need is to go through the list, check for which CVEs we have valid cache entries (based on comparing `update_dt` for equality), and process the missing ones (and cache them for subsequent queries, which may have different parameters).

kdudka (changes requested):

src/aegis_ai_web/src/endpoints/bot_kpi.py
kdudka: @superbuggy Actually, I can see that `_timestamp_in_range()` is used for processing of the data (rather than caching).  Let me take another look...

kdudka: Right, this is used for the Aegis suggestion timestamps, which is correct.  Sorry for the noise!

kdudka (commented):
@superbuggy Looks good.  Please rebase this onto `main` to resolve the CI failure.

kdudka (approved):
Looks good.  Thanks!

